PR #5, opened 2021-08-31
Remove loadbalancer version, not officially supported

bogdandrutu:


anuraaga (approved):

jpkrohling (approved):
I think it serves as an example of how to add an extra distribution. Instead of removing it entirely, can we instead add -contrib?

bogdandrutu:
-contrib is the full version and we will do that from here as well next

jpkrohling:
Alright, I'm merging this as it'll still be fresh in memory by the time we add contrib.

PR #26, opened 2021-10-06
Remove 'v' from the container image version

### LINKED ISSUE #25: Remove 'v' from the container image version
When publishing the v0.36.0 release, we noticed that the container image has the 'v' in the version as well, which breaks the pattern from previous versions. To keep compatibility with them, we should remove the 'v' from that.

jpkrohling:
Fixes #25

Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -35,12 +35,12 @@ builds:
       - goos: windows
         goarch: arm64
 docker_manifests:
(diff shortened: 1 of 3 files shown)


VineethReddy02 (approved):
LGTM!

PR #33, opened 2021-11-16
Prepare 0.39.0 release

codeboten:


Aneurysm9 (approved):

jpkrohling (changes requested):
This might be fine, but note that we seem to be using the builder v0.37.0 to build this:

https://github.com/open-telemetry/opentelemetry-collector-releases/blob/007228a546e7d0812700464e4c09b39c60179a9e/Makefile#L2

I ran goreleaser locally and it did work, but published the artifacts under my account:
https://github.com/jpkrohling/opentelemetry-collector/releases/tag/cmd/builder/v0.39.0

So, we have three options right now:

1. Release it with the 0.37.0 builder (this PR) after doing a manual check
2. Get the artifacts from my fork and manually upload to the 0.39.0 release and change this PR to use that. The builder repo has e2e tests, so, I'm sure it will work with the core 0.39.0
3. Like options 2, but run goreleaser against the main repository to get the artifacts uploaded there. A maintainer of the core is likely needed for this.

jpkrohling:
I marked my review as "request changes" so that it's not accidentally merged without verifying the binaries manually. Once that happens, feel free to mark my review as dismissed.

codeboten:
FWIW, i've manually tested the config in this repo with the builder I installed from the core repo running locally and found no problems.

jpkrohling:
Sounds good then! The only concern would have been if the bootstrap files are affected by breaking changes, namely in the `go.opentelemetry.io/collector/component` or `go.opentelemetry.io/collector/service` packages.

jpkrohling (approved):

PR #47, opened 2021-12-18
Improve the make experience.

hdost:
* Add checks for all the tools which are used by the build process.
* Switch from `which` to `commmand` since
** `command` is a builtin to the shell.
** This also fixes an issue which was causing tools to be downloaded
   reguardless of whether it has been downloaded previously.

Signed-off-by: Harold Dost <harolddost@gmail.com>


jpkrohling (approved):

hdost:
Ah actually I think it's that  didn't make a change to the GoReleaser, but that it's required for a merge to happen :/

PR #48, opened 2021-12-18
Explicitly Mention Core Components

hdost:
* Update distributions
** Add references to the core receivers, processors, extensions,
  exporters.
* Add flag `include_core` to not use implicit mentions inclusion.
* Minor indentation and spacing edits

Relates to https://github.com/open-telemetry/opentelemetry-collector/issues/3927

Signed-off-by: Harold Dost <harolddost@gmail.com>

### DIFF
--- a/distributions/otelcol-contrib/manifest.yaml
+++ b/distributions/otelcol-contrib/manifest.yaml
@@ -3,10 +3,15 @@ dist:
   name: otelcol-contrib
   description: OpenTelemetry Collector Contrib
   version: 0.41.0
+  include_core: false
   output_path: ./_build
   otelcol_version: 0.41.0
 
 extensions:
+  - import: go.opentelemetry.io/collector/extension/zpagesextension
+    gomod: go.opentelemetry.io/collector v0.41.0
+  - import: go.opentelemetry.io/collector/extension/ballastextension
+    gomod: go.opentelemetry.io/collector v0.41.0
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/asapauthextension v0.41.0
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/awsproxy v0.41.0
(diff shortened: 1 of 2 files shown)


jpkrohling (approved):
LGTM, thank you for this PR!

alolita (approved):

Aneurysm9 (approved):

hdost:
Maybe temporarily the requirement for GoReleaser can be changed? Either way https://github.com/hdost/testing-github-actions/pull/1/files << Might be a way that it could be changed for the future.

jpkrohling:
> Maybe temporarily the requirement for GoReleaser can be changed?

Yes, the requirement is new and not fully tested :-) Given that there *is* a change to `.goreleaser`, it's correct that the "Check GoReleaser Configuration" workflow should be executed, no? If that was just part of a test and the actual PR won't have it, I'll disable this check permanently until we have something in place like your scratch branch.

hdost:
> > Maybe temporarily the requirement for GoReleaser can be changed?
> 
> 
> 
> Yes, the requirement is new and not fully tested :-) Given that there *is* a change to `.goreleaser`, it's correct that the "Check GoReleaser Configuration" workflow should be executed, no? If that was just part of a test and the actual PR won't have it, I'll disable this check permanently until we have something in place like your scratch branch.

I'll remove it but i think you might have to trigger the workflows since I'm not a code owner

jpkrohling:
> I'll remove it but i think you might have to trigger the workflows since I'm not a code owner

I removed the requirement, it should be fine to merge without triggering .goreleaser. It's optional because it takes a really long time (30m or so), which is not reasonable for changes not touching .goreleaser.

hdost:
Ok I've reverted the PR back to the originally approved change https://github.com/open-telemetry/opentelemetry-collector-releases/pull/48#event-5787906982

jpkrohling (approved):

PR #43, opened 2021-12-15
Add otelcol-contrib release

jpkrohling:
Fixes #37

Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

Issue #40, opened 2021-12-14
Building arm6 container image not working


jpkrohling:
After #39, running `goreleaser --snapshot --rm-dist` doesn't seem to work anymore, breaking on this step:

```
   ⨯ release failed after 53.58s error=failed to build otel/opentelemetry-collector:0.41.0-SNAPSHOT-31eb1eb-armv6: exit status 125: [1/3] STEP 1/2: FROM alpine:3.13 AS certs
Resolved "alpine" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull docker.io/library/alpine:3.13...
Getting image source signatures
Copying blob sha256:846f3a0f3493d22f22643b6a1c057d2d37e608433cd1033d25dac92032b8b9e3
Copying config sha256:11ed10a9502532148b48c67e39d24a465d67adc7e683e830e6e5f5e597ba61a5
Writing manifest to image destination
Storing signatures
[1/3] STEP 2/2: RUN apk --update add ca-certificates
exec container process `/bin/sh`: Exec format error
[2/3] STEP 1/3: FROM alpine:3.13 AS otelcol
Error: error building at STEP "RUN apk --update add ca-certificates": error while running runtime: exit status 1
```

@silverlyra, do you remember whether this worked for you?

jpkrohling:
Same for arm7:

```
   ⨯ release failed after 35.23s error=failed to build otel/opentelemetry-collector:0.41.0-SNAPSHOT-bc17ab4-armv7: exit status 125: [1/3] STEP 1/2: FROM alpine:3.13 AS certs
Resolved "alpine" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull docker.io/library/alpine:3.13...
Getting image source signatures
Copying blob sha256:767c1b67bc22761376676021ea5a310e7a7d1344b7a017bb8ede1a202340dbaf
Copying config sha256:c7140641fa494f66493a89308fdb3f8f168cb1b83b575ff04fc5a9bb6de79eaf
Writing manifest to image destination
Storing signatures
[1/3] STEP 2/2: RUN apk --update add ca-certificates
exec container process `/bin/sh`: Exec format error
[2/3] STEP 1/3: FROM alpine:3.13 AS otelcol
Error: error building at STEP "RUN apk --update add ca-certificates": error while running runtime: exit status 1
```

jpkrohling:
And `arm64` :(

```
[...]

PR #39, opened 2021-12-09
Build ARMv6, ARMv7, and ARM64 Docker images

silverlyra:
Picking up open-telemetry/opentelemetry-collector#3331, this adds adds ARM architectures to goreleaser `dockers` and includes them in the Docker manifest.

It adds duplication to `docker.template.yaml` that could be resolved with more `yq`, but I didn’t want to YOLO-rewrite the YAML merge process that’s already present.

PR #55, opened 2022-01-11
Build ARM64 Docker images

silverlyra:
#39 added support for ARM + ARM64 multi-arch builds, but in #40 a project maintainer reported no longer being able to run goreleaser builds, and reverted it.

ARM(32) builds were removed from goreleaser entirely in #43, so this PR just adds back ARM64 Docker builds. It also updates the GitHub Actions workflows to install support for building the cross-architecture image, and updates the contributing docs with notes about how to set that up on a local system.

### DIFF
--- a/.github/workflows/ci-goreleaser.yaml
+++ b/.github/workflows/ci-goreleaser.yaml
@@ -18,10 +18,20 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Set up QEMU
+        id: qemu
+        uses: docker/setup-qemu-action@v1
+        with:
+          platforms: arm64
+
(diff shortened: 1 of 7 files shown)


Aneurysm9 (approved):

.github/workflows/ci-goreleaser.yaml
```diff
@@ -18,10 +18,20 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Set up QEMU
```
jpkrohling, line 21: That was the missing piece, thanks!

jpkrohling (approved):
Should we attempt to do a patch release to test the process before 0.43.0?

Issue #49, opened 2021-12-28
Publish images to registries other than Docker Hub


ldx:
Docker Hub has introduced pretty aggressive rate limits. We're trying to migrate our deployments to use either public.ecr.aws or quay.io instead, as trying to pull from Docker Hub fails with 429s frequently.

Any chance the images built here could also be published to other registries?

jpkrohling:
Would GitHub container repository work for you? It would be trivial to implement it, as it wouldn't require any new accounts elsewhere.

ldx:
> Would GitHub container repository work for you? It would be trivial to implement it, as it wouldn't require any new accounts elsewhere.

Absolutely, that's a great option.

jpkrohling:
Would you like to give it a try, @ldx?

ldx:
> Would you like to give it a try, @ldx?

I'm not seeing any public package in the repo, is it some other repo?

jpkrohling:
Not sure I get the question, but we don't have any packages yet, this would be the first.

ldx:
> Not sure I get the question, but we don't have any packages yet, this would be the first.

OIC, sorry, I had the impression there was an image in GHCR that I could test.

PR #56, opened 2022-01-12
Use Go to generate .goreleaser.yml

silverlyra:
Currently, the `.goreleaser.yml` file is generated with a mix of bash, sed, envsubst, and yq. With #55, it also includes a copy-pasted section in order to build images for both the amd64 and arm64 architectures.

I was thinking about how to tidy up that process, and I checked out the source code for goreleaser and discovered that the config structs it reads from `.goreleaser.yml` are [in an exported Go package](https://pkg.go.dev/github.com/goreleaser/goreleaser@v1.2.5/pkg/config#Project). So this replaces `generate-goreleaser-config.sh` with a Go program that generates the goreleaser configuration.

The primary difference between the new generated `.goreleaser.yml` and the one on the main branch it it’ll now build Docker images for the same set of CPU architectures that Go builds are run for (386, amd64, and arm64).


Aneurysm9 (approved):
This is great! I got tripped up during the last release with some of the `sed` use not being portable, so happy to see this moved to Go.  It's also a lot easier to understand since we're all Go developers.

jpkrohling:
I'll review this soon, but wanted to share that this is a great idea. I hacked those scripts together with the idea of eventually refactoring once we had some stability and knew what we needed. I think we are at that point now.

Makefile
```diff
@@ -46,29 +46,20 @@ else
...
-	[ "$${machine}" != x86 ] || machine=386 ;\
-	[ "$${machine}" != x86_64 ] || machine=amd64 ;\
-	echo "Installing yq ($${os}/$${machine}) at $(YQ_DIR)";\
-	mkdir -p $(YQ_DIR) ;\
-	curl -sLo $(YQ) https://github.com/mikefarah/yq/releases/download/v$(YQ_VERSION)/yq_$${os}_$${machine} ;\
-	chmod +x $(YQ) ;\
+		if ! command -v '$(GORELEASER)' >/dev/null 2>/dev/null; then \
+			echo >&2 '$(GORELEASER) command not found. Please install goreleaser. https://goreleaser.com/install/'; \
+			exit 1; \
+		fi \
```
silverlyra, line 64: I tried to adapt the existing pattern of the `go:` rule, but ran into an issue. The Makefile is structured so that only targets that use a tool will depend on the tool’s rule, but the `ifeq` and `error` are evaluated unconditionally by make as the Makefile is loaded.

So I ended up rewriting the `go` target and no `goreleaser` target as these shell snippets.

goreleaser/configure.go
```diff
@@ -0,0 +1,223 @@
...
+	"path"
+	"strings"
+
+	"github.com/goreleaser/goreleaser/pkg/config"
+	"github.com/goreleaser/nfpm/v2/files"
+	yaml "gopkg.in/yaml.v2"
+)
+
+var (
+	ImagePrefixes = []string{"otel"}
```
silverlyra, line 24: I made this a slice in order to pave the way for #49. We’ll be able to publish the images to additional registries by just adding them here (and adding their authentication to CI).

silverlyra (commented):
@jpkrohling PTAL when you’d like; this now [passes goreleaser CI](https://github.com/open-telemetry/opentelemetry-collector-releases/runs/4794420065?check_suite_focus=true).

The `.goreleaser.yml` diff is hard to interpret because everything is moved around; I normalized the files with `yq` and `jq` and did some manual moving around of list items to get a diff of actual changes (🤞):

```diff
--- main/.goreleaser.yml	2022-01-12 12:37:58.004874103 -0800
+++ go/.goreleaser.yml	2022-01-12 12:33:55.388176655 -0800
@@ -15,7 +15,7 @@
     flags:
       - -trimpath
     goarch:
-      - 386
+      - "386"
       - amd64
       - arm64
     goos:
@@ -24,12 +24,11 @@
       - windows
     id: otelcol
     ignore:
-      - goarch: arm
-        goos: windows
       - goarch: arm64
         goos: windows
     ldflags:
-      - -s -w
+      - -s
+      - -w
   - binary: otelcol-contrib
     dir: distributions/otelcol-contrib/_build
     env:
@@ -37,7 +36,7 @@
     flags:
       - -trimpath
     goarch:
-      - 386
+      - "386"
       - amd64
       - arm64
     goos:
@@ -46,26 +45,43 @@
       - windows
     id: otelcol-contrib
     ignore:
-      - goarch: arm
-        goos: windows
       - goarch: arm64
         goos: windows
     ldflags:
-      - -s -w
+      - -s
+      - -w
 checksum:
   name_template: '{{ .ProjectName }}_checksums.txt'
 docker_manifests:
   - image_templates:
+      - otel/opentelemetry-collector:{{ .Version }}-386
       - otel/opentelemetry-collector:{{ .Version }}-amd64
       - otel/opentelemetry-collector:{{ .Version }}-arm64
     name_template: otel/opentelemetry-collector:{{ .Version }}
   - image_templates:
+      - otel/opentelemetry-collector-contrib:{{ .Version }}-386
       - otel/opentelemetry-collector-contrib:{{ .Version }}-amd64
       - otel/opentelemetry-collector-contrib:{{ .Version }}-arm64
     name_template: otel/opentelemetry-collector-contrib:{{ .Version }}
 dockers:
   - build_flag_templates:
       - --pull
+      - --platform=linux/386
+      - --label=org.opencontainers.image.created={{.Date}}
+      - --label=org.opencontainers.image.name={{.ProjectName}}
+      - --label=org.opencontainers.image.revision={{.FullCommit}}
+      - --label=org.opencontainers.image.version={{.Version}}
+      - --label=org.opencontainers.image.source={{.GitURL}}
+    dockerfile: distributions/otelcol/Dockerfile
+    extra_files:
+      - configs/otelcol.yaml
+    goarch: "386"
+    goos: linux
+    image_templates:
+      - otel/opentelemetry-collector:{{ .Version }}-386
+    use: buildx
+  - build_flag_templates:
+      - --pull
       - --platform=linux/amd64
       - --label=org.opencontainers.image.created={{.Date}}
       - --label=org.opencontainers.image.name={{.ProjectName}}
@@ -76,6 +92,7 @@
     extra_files:
       - configs/otelcol.yaml
     goarch: amd64
+    goos: linux
     image_templates:
       - otel/opentelemetry-collector:{{ .Version }}-amd64
     use: buildx
@@ -91,11 +108,28 @@
     extra_files:
       - configs/otelcol.yaml
     goarch: arm64
+    goos: linux
     image_templates:
       - otel/opentelemetry-collector:{{ .Version }}-arm64
     use: buildx
   - build_flag_templates:
       - --pull
+      - --platform=linux/386
+      - --label=org.opencontainers.image.created={{.Date}}
+      - --label=org.opencontainers.image.name={{.ProjectName}}
+      - --label=org.opencontainers.image.revision={{.FullCommit}}
+      - --label=org.opencontainers.image.version={{.Version}}
+      - --label=org.opencontainers.image.source={{.GitURL}}
+    dockerfile: distributions/otelcol-contrib/Dockerfile
+    extra_files:
+      - configs/otelcol-contrib.yaml
+    goarch: "386"
+    goos: linux
+    image_templates:
+      - otel/opentelemetry-collector-contrib:{{ .Version }}-386
+    use: buildx
+  - build_flag_templates:
+      - --pull
       - --platform=linux/amd64
       - --label=org.opencontainers.image.created={{.Date}}
       - --label=org.opencontainers.image.name={{.ProjectName}}
@@ -106,6 +140,7 @@
     extra_files:
       - configs/otelcol-contrib.yaml
     goarch: amd64
+    goos: linux
     image_templates:
       - otel/opentelemetry-collector-contrib:{{ .Version }}-amd64
     use: buildx
@@ -121,6 +156,7 @@
     extra_files:
       - configs/otelcol-contrib.yaml
     goarch: arm64
+    goos: linux
     image_templates:
       - otel/opentelemetry-collector-contrib:{{ .Version }}-arm64
     use: buildx
```

Makefile
```diff
@@ -1,12 +1,10 @@
-GO=$(shell which go)
+GO ?= go
```
jpkrohling, line 1: IIRC there was a problem by just calling `go`: the one from the outer build (set by the github action) might be different than the one from the inner build (the one called by the builder to compile the distribution). This is typically the case when using a different go version than the default one. Using `which go` will get the path for the Go at the outer build and use that in the inner build.

distributions/otelcol-contrib/distribution.conf
```diff
@@ -1 +0,0 @@
-CONTAINER_BASE_NAME=otel/opentelemetry-collector-contrib
```
jpkrohling, line 1: I'm glad to see this file go away.

goreleaser/configure.go
```diff
@@ -0,0 +1,223 @@
...
+
+		Goos:   []string{"darwin", "linux", "windows"},
+		Goarch: Architectures,
+		Ignore: []config.IgnoredBuild{
+			{Goos: "windows", Goarch: "arm64"},
+		},
+	}
+}
+
+func Archives(dists []string) (r []config.Archive) {
```
jpkrohling, line 86: Looks good for now, but for the future, should we consider having each goreleaser section in its own source file?

jpkrohling (commented):
This looks great and I'd be OK in merging this as is. Before doing that though, do you think it would be too hard to generate a release in your own namespace, to try the whole workflow? I remember doing that in earlier versions, but I think the Docker steps weren't part of the recipe yet.

If it's too much trouble, we can try to have an unannounced patch release instead.

Makefile
silverlyra: I’ve been scratching my head, and I’m pretty sure this is a non-issue. There’s nothing in the Makefile that modifies `PATH`; running `go` from within `build.sh` will run the same binary that `command -v` would show in the Makefile.

But I’ve updated the `echo`s in `build.sh` to show (e.g.) `echo "Using Go: $(command -v "$GO")"` so the log output does include the fully-resolved path. (`which` is [less portable](https://hynek.me/til/which-not-posix/).)

I did notice, however, that `GO` doesn’t apply to the `goreleaser-verify` target. It looks like the only way to override which toolchain goreleaser uses is [in the YAML](https://github.com/goreleaser/goreleaser/blob/main/internal/builders/golang/build.go#L38-L40).

goreleaser/configure.go
silverlyra: iirc, `package main` cannot be split into multiple files when using `go run`. But the sections could be in a non-main imported package.

However, unless/until there’s a big increase in complexity in the generation code, I think it’s better this way anyway. At least, speaking for myself as someone who recently had to figure out how `.goreleaser.yml` was generated, I struggled a bit with all the small YAML snippets in different files.

It’d still have been easier to follow in multiple Go files, with top-level construction of `config.Project` being implicit vs. `yq` merging, and the ability to jump to definitions, but at ~200 lines of code I don’t think it’s worth it. Here the steps are laid out in the same order that goreleaser performs them – you can look at the CI build log and skim this file and understand quickly exactly how the setup works.

silverlyra:
@jpkrohling I configured the `DOCKER` secrets in my fork, changed `otel` to my Docker Hub username, and pushed a `v0.42.1` tag.

The release pipeline [succeeded](https://github.com/silverlyra/opentelemetry-collector-releases/runs/4798991294?check_suite_focus=true), and pushed [a GitHub release](https://github.com/silverlyra/opentelemetry-collector-releases/releases) as well as Docker images for both [the core collector](https://hub.docker.com/r/silverlyra/opentelemetry-collector/tags) and [contrib](https://hub.docker.com/r/silverlyra/opentelemetry-collector-contrib/tags). 🚢

jpkrohling:
Thank you!

PR #67, opened 2022-01-28
Add k8seventsreceiver to contrib

hprajapati-splunk:


codeboten (commented):
It looks like this component has been enabled: https://github.com/open-telemetry/opentelemetry-collector-releases/blob/6d05dbdce858c618724f8e9c470d283d58a837c2/distributions/otelcol-contrib/manifest.yaml#L121

PR #71, opened 2022-02-16
Prepare v0.45.0

jpkrohling:
Closes https://github.com/open-telemetry/opentelemetry-collector/issues/4875
Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>


tigrannajaryan (approved):

bogdandrutu (approved):

.gitignore
```diff
@@ -2,3 +2,4 @@ vendor
 **/_build
 dist/
 .idea
+.generated-yaml/
```
bogdandrutu, line 5: why this change?

jpkrohling: When testing it locally, I noticed that this directory existed and was added to my commit. I think it's not needed anymore after we replaced shell scripts with Go, but there might be workspaces out there with it.

PR #76, opened 2022-02-20
Keep latest Docker image tag up to date

### LINKED ISSUE #73: `latest` tag not updated
The [latest](https://hub.docker.com/layers/otel/opentelemetry-collector/latest/images/sha256-cec18c72535bebeaca3615e841b19066b468f87a2e29b724b118b5b7bf8808ce?context=explore) tag in dockerhub isn't getting updated during the release process. This is causing user confusion.

silverlyra:
This adds [an action][lia] to the end of the release workflow. After goreleaser pushes the Docker images, this action will semver-compare the newly-pushed image version against `latest`. If the new version is newer (and not a prerelease), it overwrites `latest` with the new manifest.

I’d been struggling to find a tidy way to do this, and ended up writing this GitHub Action just to do this one thing. 🙃

[lia]: https://github.com/silverlyra/latest-image-action

Fixes #73.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -33,8 +33,11 @@ jobs:
       - name: Generate distribution sources
         run: make generate-sources
 
-      - name: Log into Docker.io
-        run: echo "${{ secrets.DOCKER_PASSWORD }}" | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1.12.0
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
 
       - name: Login to GitHub Package Registry
         uses: docker/login-action@v1.12.0
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -53,3 +53,13 @@ jobs:
           args: release --rm-dist --timeout 1h
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Update latest tag
+        uses: silverlyra/latest-image-action@v0.1.0
+        with:
+          repository: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -50,3 +53,13 @@ jobs:
           args: release --rm-dist --timeout 1h
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Update latest tag
+        uses: silverlyra/latest-image-action@v0.1.0
```
jpkrohling, line 58: I'm a bit awry in using this as I have long-term maintenance concerns, but I think it's OK for now until we think about a more permanent solution.

.github/workflows/release.yaml
```diff
@@ -50,3 +53,13 @@ jobs:
...
+        with:
+          repository: |
+            otel/opentelemetry-collector
+            otel/opentelemetry-collector-contib
```
jpkrohling, line 62: s/contib/contrib

.github/workflows/release.yaml
```diff
@@ -50,3 +53,13 @@ jobs:
...
+            ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector
+            ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contib
```
jpkrohling, line 64: s/contib/contrib

jpkrohling:
Sorry for taking so long to review this, but better late than never, I suppose :-)

.github/workflows/release.yaml
codeboten, line 62:
```suggestion
            otel/opentelemetry-collector-contrib
```

.github/workflows/release.yaml
codeboten, line 64:
```suggestion
            ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib
```

PR #83, opened 2022-03-17
Exclude 386 from opentelemetry-collector-contrib builds

mx-psi:
Context: https://github.com/open-telemetry/opentelemetry-collector/issues/5013#issuecomment-1070762107

We remove both Linux and Windows, since the issue affects all OSs.

### DIFF
--- a/goreleaser/configure.go
+++ b/goreleaser/configure.go
@@ -22,10 +22,17 @@ import (
 
 var (
 	ImagePrefixes = []string{"otel", "ghcr.io/open-telemetry/opentelemetry-collector-releases"}
-	Architectures = []string{"386", "amd64", "arm64"}
 
 	distsFlag = flag.String("d", "", "Collector distributions(s) to build, comma-separated")
 )
+
+func architecturesForDist(dist string) []string {
+	architectures := []string{"386", "amd64", "arm64"}
+	if dist == "otelcol-contrib" {
+		architectures = []string{"amd64", "arm64"}
+	}
+	return architectures
+}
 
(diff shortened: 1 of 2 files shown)


mx-psi:
cc @jpkrohling can you take a look?

goreleaser/configure.go
```diff
@@ -22,11 +22,18 @@ import (
...
 var (
 	ImagePrefixes = []string{"otel", "ghcr.io/open-telemetry/opentelemetry-collector-releases"}
-	Architectures = []string{"386", "amd64", "arm64"}
 
 	distsFlag = flag.String("d", "", "Collector distributions(s) to build, comma-separated")
 )
 
+func architecturesForDist(dist string) []string {
+	architectures := []string{"386", "amd64", "arm64"}
+	if dist == "otelcol-contrib" {
```
jpkrohling, line 31: This is a good solution for this release and for the time being, but this would need to go into a config somewhere. I would even argue that we can remove 386 entirely, not only for the contrib. But we can have this discussion after this release :-)

jpkrohling (approved):

goreleaser/configure.go
mx-psi: IMO reintroducing 386 is fine (since the additional effort to support it is small) but we need to have CI tests first on core and contrib to prevent issues like the one on this release (I created open-telemetry/opentelemetry-collector/issues/5031 and open-telemetry/opentelemetry-collector-contrib/issues/8544 to track this)

PR #92, opened 2022-03-22
Add mongodb receiver to collector-contrib manifest

juiceppe:
Adding (MongoDB receiver)[https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/mongodbreceiver] to manifest

### DIFF
--- a/distributions/otelcol-contrib/manifest.yaml
+++ b/distributions/otelcol-contrib/manifest.yaml
@@ -123,6 +123,7 @@ receivers:
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kafkareceiver v0.47.0
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kubeletstatsreceiver v0.47.0
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/memcachedreceiver v0.47.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/mongodbreceiver v0.47.0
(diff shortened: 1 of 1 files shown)


jpkrohling:
@mx-psi, sanity check: this should have triggered the full build, shouldn't it?

mx-psi:
> @mx-psi, sanity check: this should have triggered the full build, shouldn't it?

Yes, it should have 🤔 didn't it?

jpkrohling:
I would expect the same checks to be executed as this PR: https://github.com/open-telemetry/opentelemetry-collector-releases/pull/83/checks

mx-psi:
> I would expect the same checks to be executed as this PR: https://github.com/open-telemetry/opentelemetry-collector-releases/pull/83/checks

I see. Maybe the leading `./` should be removed from the path filters? I don't see anything related [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) but it's true that examples don't have it.

jpkrohling:
In any case, the CI is failing for this PR.

codeboten (commented):
Please rebase so we can get this approved and merged

codeboten:
The mongodb receiver is part of the contrib distribution: https://github.com/open-telemetry/opentelemetry-collector-releases/blob/bba28fa44577890086d9f3f4df7d2b758fef0375/distributions/otelcol-contrib/manifest.yaml#L128

codeboten:
The mongodb receiver is part of the contrib distribution: https://github.com/open-telemetry/opentelemetry-collector-releases/blob/bba28fa44577890086d9f3f4df7d2b758fef0375/distributions/otelcol-contrib/manifest.yaml#L128

PR #144, opened 2022-06-01
Revert "Exclude 386 from opentelemetry-collector-contrib builds"

mx-psi:
Reverts open-telemetry/opentelemetry-collector-releases#83


mx-psi:
cc @codeboten, I think we can add 386 builds now after your last PR on contrib?

PR #145, opened 2022-06-03
push latest tag for docker images

### LINKED ISSUE #73: `latest` tag not updated

codeboten:
This is an alternative implementation to https://github.com/open-telemetry/opentelemetry-collector-releases/pull/76. I'll leave it up to reviewers on whether the bash script is easier to maintain than a github action, I don't have a strong preference here, but am not very familiar with typescript which is why i put together this alternative PR.

Fix #73

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -48,3 +48,10 @@ jobs:
           args: release --rm-dist --timeout 1h
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Push latest Docker tag
+        run: |
+          ./scripts/push-latest.sh otel/opentelemetry-collector ${{ github.ref_name }}
+          ./scripts/push-latest.sh otel/opentelemetry-collector-contrib ${{ github.ref_name }}
+          ./scripts/push-latest.sh ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector ${{ github.ref_name }}
+          ./scripts/push-latest.sh ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contib ${{ github.ref_name }}
--- a/scripts/push-latest.sh
+++ b/scripts/push-latest.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+image=$1
+candidate=$2
+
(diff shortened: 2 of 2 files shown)


bogdandrutu (commented):
How can we test this?

codeboten:
> How can we test this?

I ran some test of the script locally, I can try to see if I can run the tests against a fork of this repo

.github/workflows/release.yaml
```diff
@@ -48,3 +48,10 @@ jobs:
           args: release --rm-dist --timeout 1h
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Push latest Docker tag
+        run: |
+          ./scripts/push-latest.sh otel/opentelemetry-collector ${{ github.ref_name }}
```
jpkrohling, line 54: Isn't there a way to add this to goreleaser's config? Like here:

https://github.com/open-telemetry/opentelemetry-collector-releases/blob/1af799780977b1ef24d628ed3665a64d704f6263/.goreleaser.yaml#L235-L239

mx-psi: Based on https://goreleaser.com/customization/docker/#customization I think we can just add `otel/opentelemetry-collector:latest` in the appropriate place?

jpkrohling: Yes, but note that the "docker" section is for the arch-specific images, whereas the "docker_manifests" is the meta image holding the arch-specific ones. The meta one is the one people pull most of the time.

mx-psi (changes requested):
This could be quite disruptive to users if they are not aware of the current situation, since they may face a lot of breaking changes.

We should do this only for `otel/opentelemetry-collector` (and the associated one on the ghcr.io registry) on this first PR, evaluate the damage done, and do contrib on a different release.

codeboten:
superseeded by https://github.com/open-telemetry/opentelemetry-collector-releases/pull/220

PR #154, opened 2022-06-29
Add google managed prometheus exporter to otelcol-contrib

damemi:
Ref https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.54.0/exporter/googlemanagedprometheusexporter


damemi (commented):
cc @dashpole fyi

PR #164, opened 2022-07-14
[docs] add descriptions for distros

### LINKED ISSUE #121: Add definition for distributions
Currently the list of distributions only link to the contents of the directories. It would be useful to describe the intent of each distribution.

codeboten:
As the distributions are linked from various places, it makes sense to provide end-users a bit more clarity on what each distro includes.

Fixes #121

### DIFF
--- a/distributions/otelcol-contrib/README.md
+++ b/distributions/otelcol-contrib/README.md
@@ -9,7 +9,7 @@ As this distribution contains many components, it is a good starting point to tr
 * reduce the size of the collector, reducing deployment times for the collector
 * improve the security of the collector by reducing the available attack surface area
 
-Building a custom collector can be achieved using the [OpenTelemetry Collector Builder](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder).
+Building a [custom collector](https://opentelemetry.io/docs/collector/custom-collector/) can be achieved using the [OpenTelemetry Collector Builder](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder).
 
 ## Components
 
--- a/distributions/otelcol-contrib/README.md
+++ b/distributions/otelcol-contrib/README.md
@@ -1,0 +1,16 @@
+# OpenTelemetry Collector Contrib Distro
+
+This distribution contains all the components from both the [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector) repository and the [OpenTelemetry Collector Contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib) repository. This distribution includes open source and vendor supported components.
+
+## Recommendation
+
+As this distribution contains many components, it is a good starting point to try various configurations. However, when running in production, it is recommended to limit the collector to contain only the components necessary for an environment. Some reasons to do this:
+
+* reduce the size of the collector, reducing deployment times for the collector
+* improve the security of the collector by reducing the available attack surface area
+
+Building a custom collector can be achieved using the [OpenTelemetry Collector Builder](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder).
+
+## Components
+
+The full list of components is available in the [manifest](manifest.yaml)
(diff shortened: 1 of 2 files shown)


distributions/otelcol-contrib/README.md
```diff
@@ -0,0 +1,16 @@
+# OpenTelemetry Collector Contrib Distro
+
+This distribution contains all the components from both the [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector) repository and the [OpenTelemetry Collector Contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib) repository. This distribution includes open source and vendor supported components.
+
+## Recommendation
+
+As this distribution contains many components, it is a good starting point to try various configurations. However, when running in production, it is recommended to limit the collector to contain only the components necessary for an environment. Some reasons to do this:
+
+* reduce the size of the collector, reducing deployment times for the collector
+* improve the security of the collector by reducing the available attack surface area
```
Aneurysm9, line 10: Is this true?  I suppose components could have an `init()` function that did something malicious or opened an avenue for interaction, but in the normal course of things more components only means more factory functions in the service's factory map.  It's not until a component is used in a configuration that it is instantiated and could become a security concern.

TylerHelmuth: I think you're right @Aneurysm9, but I believe the sentiment here is "more components means more opportunity for a compromised component, which means more opportunity that you'll use a compromised component".  I think its all about increased chances that something could happen due to the large number of components that you could chose to use.

codeboten: @Aneurysm9 in theory if someone enabled remote configuration (as OpAmp progresses) I would expect this to be true, I'm happy to remove this if you disagree.

jpkrohling: Vulnerabilities are typically exploited by taking advantage of multiple factors. Having all possible components in there means that we are making an attacker's life easier. It's good practice to reduce the surface attack to only the things that are absolutely necessary.

jpkrohling (approved):

distributions/otelcol-contrib/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+Building a custom collector can be achieved using the [OpenTelemetry Collector Builder](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder).
```
mx-psi, line 12: Is the builder itself ready for production usage? It definitely does the job, since we use it for official releases, but 
- we don't have docs about it on https://opentelemetry.io
- we don't have 'ocb' Docker images
- people need to add `replace` statements to actually make the thing build, which we don't document anywhere AFAICT

I am not opposed in principle to this PR, but I think we need to improve our docs and packaging for this to be a feasible solution for many people, and also make sure that we are promoting usage of something that is mature enough

mx-psi: I opened open-telemetry/opentelemetry-collector/issues/5713 and open-telemetry/opentelemetry-collector/issues/5712 to track a couple of these.

TylerHelmuth: > we don't have docs about it on https://opentelemetry.io/

There is this doc: https://opentelemetry.io/docs/collector/distributions/#custom-distributions

mx-psi: > There is this doc: https://opentelemetry.io/docs/collector/distributions/#custom-distributions

There's actually this doc https://opentelemetry.io/docs/collector/custom-collector/ I was wrong about that one :) I think the other points still apply though

TylerHelmuth: > I think the other points still apply though

I agree

TylerHelmuth: I am good with moving this forward, but it also matches my opinions on contrib's distro haha

TylerHelmuth (approved):

distributions/otelcol-contrib/README.md
mx-psi: I am good with moving forward too, I opened the issues a while back so this works for me :)

mx-psi (approved):

Aneurysm9 (approved):

PR #179, opened 2022-08-02
Copy etc/os-release to final images to make vuln scanners to work

### LINKED ISSUE #175: etc/os-release is not copied to final image
GCP [container analysis services](https://cloud.google.com/container-analysis/docs/container-scanning-overview#linux-versions) which perform vulnerability scans of images used in conjunction with GKE uses `etc/os-release` to determine the OS of the image.

otel collector does use Alpine 3.11 which is a supported base image, but the specific file isn't copied into the final image so the service is unable to scan it.

Could you add this step (ideally to all otel images) so that they are able to be scanned?

iblancasa:
Signed-off-by: Israel Blancas <iblancasa@gmail.com>

Closes #175

### DIFF
--- a/distributions/otelcol-contrib/Dockerfile
+++ b/distributions/otelcol-contrib/Dockerfile
@@ -12,6 +12,7 @@ FROM scratch
 ARG USER_UID=10001
 USER ${USER_UID}
 
+COPY --from=certs /etc/os-release etc/os-release
 COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 COPY --from=otelcol-contrib /otelcol-contrib /
 COPY configs/otelcol-contrib.yaml /etc/otelcol-contrib/config.yaml
--- a/distributions/otelcol/Dockerfile
+++ b/distributions/otelcol/Dockerfile
@@ -12,6 +12,7 @@ FROM scratch
 ARG USER_UID=10001
 USER ${USER_UID}
 
+COPY --from=certs /etc/os-release etc/os-release
 COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 COPY --from=otelcol /otelcol /
 COPY configs/otelcol.yaml /etc/otelcol/config.yaml


iblancasa:
I'm not sure if this is a pretty good idea since the final images are based on `scratch` and not on `alpine` as #175 says.
https://github.com/aquasecurity/trivy/issues/173#issuecomment-546673051

jpkrohling:
That's a great point. Can you check how we can get the vulnerability scanners to give us a green checkmark? What do other projects with scratch images do?

iblancasa:
I'll try to gather some ideas... but I usually saw people using `trivy fs` command to deal with this kind of situations

piotrgwiazda:
How about having another image with "*-alpline" tag created that will be based on alpine rather than on FROM scratch? The user/consumer will be free to select the image to use.

iblancasa:
@piotrgwiazda the only thing we'll add in that case, are the vulnerabilities from the base OS. I'm not sure if the ones from the execuables will be detected... since for go projects, usually those tools check the `go.mod` file.

jpkrohling:
> How about having another image with "*-alpline" tag created that will be based on alpine rather than on FROM scratch? The user/consumer will be free to select the image to use.

This generates an extra load on our side, as we would have to support 2x the number of images we currently have. I'd argue that we should have a way to tell those tools that we only have a binary on the image, instead of an OS with other tools.

iblancasa:
The ideal would be to check the source code from each repository... but since this can be slow, I can add a script that checks out the different repositories and checks for vulnerabilities.

What do you think?

jpkrohling:
GitHub already provides that for us, it can scan our own code via CodeQL and will advise us in case a dependency has a known vulnerability. We need existing vulnerability scanners to accept our images as safe even though they can't detect the underlying OS (because we are using images from scratch).

iblancasa:
I see. Is changing the base images an option? Otherwise I think there is not a fix for this issue.

jpkrohling:
> Is changing the base images an option?

It might be an option if there's a good reason for that. So far, I see that vulnerability scanners cannot give us a green mark because we are using scratch images. This is not the same as saying we have vulnerabilities.

iblancasa:
I know but... I think there are no many options, right?

jpkrohling:
Is there anything in the docs for GCP's tool, or for Sysdig/Snyk/Aqua on how to mark our container as scratch? Perhaps a tag that we can add to the image itself?

iblancasa:
I don't find anything like that. The only thing I found is the `--vuln-type` argument from `trivy`. For instance:
```sh
trivy i --vuln-type library otel/opentelemetry-collector:latest
```

jpkrohling:
Alright, I think we did what was reasonable around this. Unless I'm missing something crucial, I believe those tools should provide a way to be told that an image is "from scratch" and should be marked as such.

iblancasa:
Hmmm I was not able to find anything related.
I did a couple of experiments... and something seems to work, is to just create a dummy file (using touch). What do you think?

PR #188, opened 2022-08-31
Update to v0.59.0

Aneurysm9:
Signed-off-by: Anthony J Mirabella <a9@aneurysm9.com>


distributions/otelcol/manifest.yaml
```diff
@@ -8,44 +8,44 @@ dist:
```
jpkrohling, line 8: I can't comment on a line before this, but lines 5 and 7 need to be updated.

distributions/otelcol-contrib/manifest.yaml
```diff
@@ -2,166 +2,166 @@ dist:
...
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/tcplogreceiver v0.59.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/udplogreceiver v0.59.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/vcenterreceiver v0.59.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/wavefrontreceiver v0.59.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/windowsperfcountersreceiver v0.59.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/zipkinreceiver v0.59.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/zookeeperreceiver v0.59.0
 
 # When adding a replace, add a comment before it to document why it's needed and when it can be removed
 replaces:
```
jpkrohling, line 167: An exercise we run during the release is whether we still need the replaces statements here. Do you think we still need the ones we have here?

Aneurysm9: Not sure I can make a call on either of these.  The `gnostic` issue is still open and people are still reporting issues, so probably safest to leave it.  The `go-connections` link I don't have enough context to follow.  @jamesmoessis or @MovieStoreGuy may be able to provide more details.  

Since these are both versioned replacements it should be safe to leave them until the creator weighs in on their removal since the won't take effect if a newer version would be selected.

bogdandrutu (approved):

PR #190, opened 2022-09-01
[chore] Increase GoReleaser timeout to 90 minutes

TylerHelmuth:
The last 2 releases ([0.58.0](https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/2967917951/attempts/1) and [0.59.0](https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/2843445596/attempts/1)) have had timeouts that required the build to be re-run.  This PR increases the timeout to 90 minutes to try to prevent this as a short-term solution.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -45,6 +45,6 @@ jobs:
         uses: goreleaser/goreleaser-action@v3
         with:
           version: latest
-          args: release --rm-dist --timeout 1h
+          args: release --rm-dist --timeout 90m
         env:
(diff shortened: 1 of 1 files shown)


jpkrohling (approved):
This is fine, but I wonder if we can somehow optimize this process so that it doesn't take 90 clock minutes.

TylerHelmuth:
> This is fine, but I wonder if we can somehow optimize this process so that it doesn't take 90 clock minutes. 

Yes, we need a long term solution. This is definitely a short term solution to help with future releases in the meantime.

PR #204, opened 2022-09-29
Add jaegerremotesampling extension to contrib

greut:
the remote sampling has been removed from the receiver but its replacement wasn't added to the extensions, cf https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/6510

So here it is.


jpkrohling:
@frzifus, can you review this one?

frzifus (approved):
oh that makes sense, thank you @greut!

codeboten (approved):

PR #205, opened 2022-09-30
bump alpine base image to 3.16

cpanato:
- bump alpine base image to 3.16

We might consider using distroless images to reduce the surface attack and have smaller final images.
if this is the desire of the maintainers I can work on that and maybe we can use https://github.com/chainguard-images/static / https://github.com/chainguard-images/alpine-base


jpkrohling (approved):
We do use a scratch image in the final result. The alpine images are only stage builds for the final image.

PR #206, opened 2022-09-30
make sure that it uses the latest patch release available

cpanato:
- make sure that it uses the latest patch release available

### DIFF
--- a/.github/workflows/ci-goreleaser.yaml
+++ b/.github/workflows/ci-goreleaser.yaml
@@ -35,7 +35,8 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19.1
+          go-version: '1.19'
+          check-latest: true
 
       - name: Generate the sources
         run: make generate-sources
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,7 +19,8 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19.1
+          go-version: '1.19'
+          check-latest: true
 
       - name: Verify
         run: make ci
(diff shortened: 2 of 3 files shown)


.github/workflows/ci-goreleaser.yaml
```diff
@@ -35,7 +35,8 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v3
```
cpanato, line 36: this and the other actions will be good to replace that to use the git hashes to avoid any issues if the tag mutate and break or other possible security implications

jpkrohling: Not sure -- this could very well be a branch, so using "v3" would just use the latest state of that branch.

.github/workflows/ci-goreleaser.yaml
```diff
@@ -35,7 +35,8 @@ jobs:
...
         with:
-          go-version: 1.19.1
+          go-version: '1.19'
+          check-latest: true
```
jpkrohling, line 39: Isn't this being bumped by dependabot already when new versions are available? What happens here if there's a 1.20 available?

cpanato: dependabot does not bump this part, it only bumps the action itself
if 1.20 is out we need to update and push a PR

jpkrohling: Right, dependabot will only change that part. When 1.20 is out, will we get build failures, warnings, or things like that?

cpanato: no, we will still run go1.19 only if we upgrade to go1.20

jpkrohling: Thanks for the clarification!

jpkrohling (approved):

.github/workflows/ci-goreleaser.yaml
cpanato: thanks for the review

PR #219, opened 2022-10-13
Add scripts to push latest image, automate this in github action

bogdandrutu:

### DIFF
--- a/scripts/latest-tag.sh
+++ b/scripts/latest-tag.sh
@@ -1,0 +1,42 @@
+VERSION=''
+
+while getopts v: flag
+do
+    case "${flag}" in
+        v) VERSION=${OPTARG};;
+    esac
+done
+
+[[ -n "$VERSION" ]] || echo "Unknown version. Use ./scripts/latest-tag.sh -v VERSION" || exit 1
+
+# Push latest tag for otel/opentelemetry-collector-contrib
+docker pull otel/opentelemetry-collector-contrib:${VERSION}-386
+docker pull otel/opentelemetry-collector-contrib:${VERSION}-amd64
+docker pull otel/opentelemetry-collector-contrib:${VERSION}-arm64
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -61,3 +64,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           COSIGN_EXPERIMENTAL: true
+
+      - name: Push latest label for core
```
jpkrohling, line 68: I know this PR is fixing a pain we have, but I'd rather have this as part of goreleaser instead.

jpkrohling (changes requested):

.github/workflows/release.yaml
bogdandrutu: Then review https://github.com/open-telemetry/opentelemetry-collector-releases/pull/220

jpkrohling: Done ;-)

codeboten:
Superseeded by https://github.com/open-telemetry/opentelemetry-collector-releases/pull/220

bogdandrutu:
@codeboten I think the manual script is still good to have, will do a PR with only that

bogdandrutu:
Let's keep it around for some time. One use-case is that if we release `0.70.0` then do a patch release to `0.69.1` that will be getting "latest" which we need then to manually fix.

.github/workflows/release.yaml
bogdandrutu: @jpkrohling review again please :)

codeboten (commented):
I'm ok with keeping the manual script around in case it's needed. Please add a comment to the file to make it clear that's when this would be used, otherwise i'm sure i'll be going around in 3 months wondering why the hell this file is here and create a PR to delete it 🤣

PR #222, opened 2022-10-13
[chore] prepare v0.62.1

bogdandrutu:


Aneurysm9 (approved):

codeboten (approved):

jpkrohling (approved):
LGTM. I double-checked that the resulting go.mod for the contrib uses core v0.62.1, which I understand brings the required fix. This should be sufficient for contrib modules using core to benefit from the patched version.

PR #226, opened 2022-10-24
add httpcheck receiver

codeboten:
This will remain in draft until v0.63.0 release


codeboten:
Added in https://github.com/open-telemetry/opentelemetry-collector-releases/pull/227

PR #302, opened 2023-03-13
Add cloudflare reciver to contrib

dehaansa:


dehaansa:
Jumped the gun a bit here, will add this once it's hit the next release :)

Issue #323, opened 2023-04-22
User otelcol-contrib requires permission to read journald


FeltrinN:
The default user `otelcol-contrib` does not have permission to read journald, which causes the journald receiver to fail (see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/20906).

One simple fix would be to add the user to the `systemd-journal` group, but I'm not sure if there are any security implications that should be taken into account.

Another possibility would be to document this requirement in the [getting started guide](https://opentelemetry.io/docs/collector/getting-started/#linux-packaging).

PR #324, opened 2023-04-22
Add otel-contrib user to systemd-journal group

FeltrinN:
This is required by the journald receiver

### DIFF
--- a/distributions/otelcol-contrib/preinstall.sh
+++ b/distributions/otelcol-contrib/preinstall.sh
@@ -14,4 +14,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-getent passwd otelcol-contrib >/dev/null || useradd --system --user-group --no-create-home --shell /sbin/nologin otelcol-contrib
+getent passwd otelcol-contrib >/dev/null || useradd --system --user-group --groups systemd-journal --no-create-home --shell /sbin/nologin otelcol-contrib


FeltrinN:
resolves #324

codeboten (commented):
@sumo-drosiek @djaglowski as the owners of that receiver, can you take a look at this PR

codeboten:
> resolves #324

@FeltrinN is there an issue associated w/ this? The link here is this PR :)

sumo-drosiek:
@codeboten My only concern is, should `otelcol-sumo` by default to be able to read all users and system journal logs

codeboten (commented):
@sumo-drosiek just confirming you mean `otelcol-contrib` here?

FeltrinN:
@codeboten @sumo-drosiek  sorry for the confusion. This started as an [`otel-contrib` issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/20906), and when it became clear that it was not about the application but about the packaging I opened [an issue here](https://github.com/open-telemetry/opentelemetry-collector-releases/issues/323) and this PR.

For what it's worth I'm also not 100% convinced that this is the right solution, but I think this is a "product" question. If the project aims to provide an easy tool that requires little to no configuration, then something along the lines of this PR is probably desirable. If otherwise the aim is to have a baseline that users should configure according to their needs, I'd avoid adding any permission that goes beyond the bare minimum and compensate by adding documentation and/or error messages to the collector to help troubleshooting the issue.

In any case, thank you very much for all your work until now :)

sumo-drosiek:
> @sumo-drosiek just confirming you mean `otelcol-contrib` here?

Yes, sorry :sweat_smile:

djaglowski:
I'm not an expert on what users would expect if deploying the contrib collector, but my inclination would be to lead towards addressing this with documentation. There are a few places that may be helpful:
- As noted in https://github.com/open-telemetry/opentelemetry-collector-releases/issues/323, the getting started guide.
- The readme for the journald receiver.

Additionally, if possible, detecting the error within the receiver and emitting a clear error message would be a nice improvement.

mx-psi:
I would also lean towards not accepting this PR; we generally have leaned towards the "baseline that users should configure according to their needs" approach and unless we have a very clear picture of the security implications and the implications for cross-platform support of doing this, we should not change our stance here.

Additionally, regardless of whether we accept this PR, we should also to handle this at the documentation level and with more descriptive errors in the journald receiver. Downstream distros and custom distributions built with ocb will not benefit from this change and therefore need to have that information to add this permission themselves.

djaglowski:
@FeltrinN, would you be willing to propose specific documentation updates with a PR on collector-contrib?

jpkrohling:
I reviewed the PR but I agree that this would be better handled with documentation. The way this PR has it, we'll be adding permissions without knowing whether people will need it, which goes against the best security practices (like "least privilege" principle).

FeltrinN:
I think everything above makes sense, therefore closing this PR.

I'll contribute to the journald receiver documentation, as I guess the getting started guide should be as essential and minimal as possible, leaving detailed information on requirements for specific modules to the modules documentation. I'll see if I can contribute the error message too, but it will definitely go faster if somebody with better knowledge of the codebase does it :)

PR #328, opened 2023-04-28
Fix core version

djaglowski:


distributions/otelcol/manifest.yaml
```diff
@@ -2,7 +2,7 @@ dist:
   module: github.com/open-telemetry/opentelemetry-collector-releases/core
   name: otelcol
   description: OpenTelemetry Collector
-  version: 0.75.0
+  version: 0.76.1
   output_path: ./_build
   otelcol_version: 0.75.0
```
evan-bradley, line 7: Should this be updated as well?

djaglowski: Yes, thanks for catching it

evan-bradley (approved):

kovrus (approved):

Issue #341, opened 2023-05-24
:bug: v0.78.0 Linux assets are missing


darkweaver87:
On last v0.78.0 it looks like your github runner went out of disk space.
Consequently, only 11 assets were created instead of the usual 43.
Thanks :-)

codeboten:
For more info on the disk running out, see https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/5062625852/jobs/9088345803#step:22:7

codeboten:
Discussed this with other maintainers, will break the release process further. Currently the github actions that produce the artifacts are split by GOOS, will split by GOARCH and distribution as well

mpostument:
Hey team, any ETA on linux binary for 0.78.0 version? I am waiting for dynamic index in elastic exporter. But cannot upgrade yet

codeboten:
Artifacts are posted now: https://github.com/open-telemetry/opentelemetry-collector-releases/releases/tag/v0.78.0, closing the issue

PR #344, opened 2023-05-25
[chore] split goreleaser by GOARCH

codeboten:
This is to address a space limit we've run into w/ caching. Additionally, I changed the workflow to use upload/download artifact to avoid having to create many copies of the cache action for each combination of GOOS+GOARCH.

Linked issue: https://github.com/open-telemetry/opentelemetry-collector-releases/issues/341

I tested the results of upload/download artifact in a private repository.

### DIFF
--- a/.github/workflows/ci-goreleaser.yaml
+++ b/.github/workflows/ci-goreleaser.yaml
@@ -20,6 +20,7 @@ jobs:
     strategy:
       matrix:
         GOOS: [linux, windows, darwin]
+        GOARCH: ["386", amd64, arm64, ppc64le]
     runs-on: ubuntu-20.04
 
     steps:
@@ -53,4 +54,5 @@ jobs:
           args: --snapshot --clean --skip-sign --skip-sbom --timeout 2h --split
         env:
           GOOS: ${{ matrix.GOOS }}
+          GOARCH: ${{ matrix.GOARCH }}
           GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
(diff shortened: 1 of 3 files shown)


codeboten:
I think using the upload/download alone may address #341 since the jobs will no longer need to create an additional archive to cache.

.github/workflows/ci-goreleaser.yaml
```diff
@@ -20,6 +20,7 @@ jobs:
     strategy:
       matrix:
         GOOS: [linux, windows, darwin]
+        GOARCH: ["386", amd64, arm64, ppc64le]
```
mx-psi, line 23: We ignore a couple combinations in the code that generates the goreleaser configuration https://github.com/open-telemetry/opentelemetry-collector-releases/blob/3971091c01a366e0b33df39a105daa2f38bef44d/cmd/goreleaser/internal/configure.go#L72-L75

Should we also ignore them in the matrix definition here?

.github/workflows/release.yaml
```diff
@@ -9,6 +9,7 @@ jobs:
     strategy:
       matrix:
         GOOS: [linux, windows, darwin]
+        GOARCH: ["386", amd64, arm64, ppc64le]
```
mx-psi, line 12: ditto, should we exclude unused combinations?

.github/workflows/ci-goreleaser.yaml
codeboten: those end up getting ignored here: https://github.com/open-telemetry/opentelemetry-collector-releases/blob/49c07b47e3154a01604622a40256c4e4b0d33de8/.goreleaser.yaml#L37-L41, but i can add an ignore to the matrix definition

codeboten: excluded 👍🏻

mx-psi (approved):

PR #386, opened 2023-08-11
Replace datadog-agent dependecy versions

songy23:
Pins datadog-agent dependency versions to `v0.48.0-beta.1` to adapt breaking changes. Same as https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/25102.


songy23:
Needs to wait for v0.83. cc @mx-psi

songy23:
@djaglowski IIUC this manifest file is for users to try out contrib components, they may still need the `replace` statement or they can run into dependency issues. I think merging this PR to mainline should be enough, I don't see a need to cut a patch release unless I'm missing something.

djaglowski (approved):

PR #476, opened 2024-02-09
Update storage modules

djaglowski:
Follows https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31040


djaglowski:
I believe this change will need to be rebased after v0.95.0 is tagged on the contrib repo, but before releasing here. I'm tagging this as a release blocker and cc @jpkrohling as the next release manager.

jpkrohling (approved):

Issue #429, opened 2023-11-21
Allow independent distribution releases


TylerHelmuth:
Our current release problem has a couple problems:

1. We cannot release distributions independently.  If we want to release a patch for the Contrib Distribution we have to also release a patch for the Core distribution.
    - Related, if the job fails after 1 distribution is published but not the other, we can't rerun the job, we have to push new tags.
2. We have run into diskspace issues in the past where the job fills up the allotted amount. The more we add to contrib and the more distributions we create the more this will be an issue.

As we look to support more distributions the connectedness of our CI will be a roadblock.

I propose we split the release of each distribution into it's own workflow with their own tags, such as `v0.90.0-contrib` or `v0.90.0-core`. The logic of each workflow can reuse a shared action, but I believe we'd need a goreleaser.yaml per distribution.

If each distribution has its own workflow we can:
1. Release each distribution independently via their own tags. If necessary we can add a make target to simplify the Release Manager's steps.
2. Each distribution will have more disk space (or any resource, including time) to work with.

Based on https://github.com/open-telemetry/opentelemetry-collector-releases/issues/357#issuecomment-1593537385.

jpkrohling:
I agree with that. I've been experimenting with splitting the distributions on their own workflows as well on a personal repository, and I prefer that approach than what I originally built in this repo here.

While I think it's great to have the ability to release different distributions separately, I don't think it's a good idea to do so: I think it might be confusing for users to have, say, otelcol-contrib-v1.0.1 and not an otelcol-core-v1.0.1 (and vice-versa).

> Related, if the job fails after 1 distribution is published but not the other, we can't rerun the job, we have to push new tags.

Care to expand on this part? Is this only about the step where tags are pushed, or is it something else? I think the release should already be idempotent for most steps.

codeboten:
When there's an issue with a release, we delete the tags and push them again. There's a note about tags in the README:

> ⚠️ Important note: Git tags in this repository may change at any time to fix any issues found during a release. They are only meant to trigger Github releases and should not be relied upon.

I agree that there needs to be separate workflows for the distros, otherwise we're limited in what we can release in the future. We've already seen this in the past.

Re. different versions for different distros, I'm not sure I agree that we should always keep the different versions in sync. I'm thinking of the world of the future where most of the components in core are stable and no longer need to release on a frequent cadence. This would mean that we could end up in a scenario where core and contrib are at different versions and that should be perfectly acceptable.

[...]

PR #477, opened 2024-02-14
[chore] Split core and contrib workflows

TylerHelmuth:
This PR separates the Core and Contrib distributions into separate workflows.

The repository continues to use a single goreleaser.yaml file because:
1. we already have a process to generate it
2. goreleaser has features to support a single file

We specifically take advantage of `id` to specify exactly which things in the goreleaser.yaml file get built.

For now I did not change the triggers - both sets of workflows will currently always be triggered at the same time, like they are today.

Related to https://github.com/open-telemetry/opentelemetry-collector-releases/issues/429

Example workflows: 
- https://github.com/TylerHelmuth/opentelemetry-collector-releases/actions/runs/7908747507
- https://github.com/TylerHelmuth/opentelemetry-collector-releases/actions/runs/7908747499

I couldn't test the actual release bc I don't have a goreleaser-pro key.


TylerHelmuth:
Created https://github.com/open-telemetry/opentelemetry-collector-releases/pull/478 to take care of ensuring `partial` is included in the generate.

.github/workflows/ci-goreleaser-core.yaml
```diff
@@ -0,0 +1,69 @@
+name: Continuous Integration - GoReleaser - Core
```
jpkrohling, line 1: I did some similar work in this repo, with reusable workflows: https://github.com/jpkrohling/otelcol-distributions

Basically, I defined the generic part here: https://github.com/jpkrohling/otelcol-distributions/blob/main/.github/workflows/base-release.yaml

And then each distribution would just pass parameters:
- https://github.com/jpkrohling/otelcol-distributions/blob/main/.github/workflows/release-loadbalancing.yaml
- https://github.com/jpkrohling/otelcol-distributions/blob/main/.github/workflows/release-sidecar.yaml
- https://github.com/jpkrohling/otelcol-distributions/blob/main/.github/workflows/release-tracing.yaml

TylerHelmuth: @jpkrohling perfect, I can implement this reusable pattern now instead of later

TylerHelmuth: Done

.github/workflows/base-release.yaml
```diff
@@ -133,8 +135,8 @@ jobs:
         with:
           distribution: goreleaser-pro
           version: latest
-          args: continue --merge --timeout 2h
+          args: continue --merge --timeout 2h --id ${{ inputs.distribution }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          COSIGN_YES: true
```
jpkrohling, line 139: Can you double-check this? I think my version is older than what we have in this repo here when it comes to cosign.

.github/workflows/base-release.yaml
```diff
@@ -98,19 +100,19 @@ jobs:
...
       - uses: docker/setup-qemu-action@v3
         with:
           platforms: arm64,ppc64le,s390x
 
       - uses: docker/setup-buildx-action@v3
 
-      - uses: actions/setup-go@v5
+      - uses: actions/setup-go@v4
         with:
-          go-version: '~1.21.7'
```
jpkrohling, line 113: Can you double-check this change? Is this downgrade intentional?

.github/workflows/base-release.yaml
```diff
@@ -98,19 +100,19 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: sigstore/cosign-installer@v3
```
jpkrohling, line 101: Can you check these downgrades? Were they intentional? Perhaps my version is older than that we have here in this repo

.github/workflows/base-release.yaml
```diff
@@ -70,12 +72,12 @@ jobs:
         with:
           distribution: goreleaser-pro
           version: latest
-          args: release --clean --split --timeout 2h
+          args: release --clean --split --timeout 2h --id $ {{ inputs.distribution }}
         env:
           GOOS: ${{ matrix.GOOS }}
           GOARCH: ${{ matrix.GOARCH }}
           GITHUB_TOKEN: ${{ secrets.GH_PAT }}
-          COSIGN_YES: true
```
jpkrohling, line 78: This also needs double-checking

jpkrohling (commented):
It looks mostly right, but there are a few cosign-related changes that seem misplaced.

TylerHelmuth:
@jpkrohling all those downgrades were the result of a bad merge.  Should now be fixed.

jpkrohling (approved):

PR #485, opened 2024-02-21
Split goreleaser config file

TylerHelmuth:
This PR adds a distribution-specific `.goreleaser.yaml` file to each distribution and updates the ci to use these specific file.

In the next PR I'll remove the auto-generated code, update the makefile, and remove the `.goreleaser.yaml` file that is at the root.

Related to https://github.com/open-telemetry/opentelemetry-collector-releases/issues/429


distributions/otelcol-contrib/.goreleaser.yaml
```diff
@@ -0,0 +1,257 @@
+partial:
+  by: target
+project_name: opentelemetry-collector-releases
```
jpkrohling, line 3: Can you change this name? This should solve the problem we had during the last release.

cc @cpanato

jpkrohling (commented):
Before approving, let's discuss whether we are ready to remove the Go code that generates the goreleaser files: removing that code generator means we'll have to maintain those two YAML files. Adding cosign, for instance, would have required adding the same things on the two files.

I'm fine if we have a reason not to generate the goreleaser files, such as significant differences between the distributions.

distributions/otelcol-contrib/.goreleaser.yaml
cpanato: looks like yes because then the checksum will be with this name

```
name_template: '{{ .ProjectName }}_checksums.txt'
```

cpanato:
lgtm

TylerHelmuth:
@jpkrohling i think we could do the same approach you've done in your repo and auto generate the separate files. I was planning to do that in a future PR.

jpkrohling (approved):

distributions/otelcol-contrib/.goreleaser.yaml
TylerHelmuth: Instead of updating the project name I went with updating the checksum directly.  `.ProjectName` is used other places and including the distribution name didnt feel right.

jpkrohling:
Can you confirm how to run this locally? I tried the following, but it failed:

```
> goreleaser --snapshot --clean 
  • starting release...
  ⨯ release failed after 0s                  error=yaml: unmarshal errors:
  line 1: field partial not found in type config.Project
```

I also tried something closer to what we have in the CI files, but also failed:

```
> goreleaser --snapshot --clean --skip-sign --skip-sbom --timeout 2h --split --id otelcol
  ⨯ command failed                                   error=unknown flag: --split
```

cpanato:
if you are using the more up-to-date goreleaser you can update the --skip flags to use `--skip=sign,sbom`

TylerHelmuth:
@jpkrohling I've never tried to run locally.  We use goreleaser-pro in the repo and I don't have access locally.  A lot of the features we use are pro features.

jpkrohling (commented):
There's still a .goreleaser at the root of the repository and needs to be removed.

.github/workflows/base-ci-goreleaser.yaml
```diff
@@ -57,7 +57,8 @@ jobs:
         with:
           distribution: goreleaser-pro
           version: latest
-          args: --snapshot --clean --skip-sign --skip-sbom --timeout 2h --split --id ${{ inputs.distribution }}
+          workdir: distributions/${{ inputs.distribution }}
+          args: --snapshot --clean --skip-sign --skip-sbom --timeout 2h --split
```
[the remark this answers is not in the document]
TylerHelmuth: @jpkrohling in this situation I feel ok with the skips.  That particular log is for linux_amd64, and it does say

```
    • building docker image                          image=otel/opentelemetry-collector:0.95.0-SNAPSHOT-a5b01f6-amd64
```

which indicates that it built the linux amd64 image.  All other images are skipped since they dont match this matrix instance's os and architecture.

Can you confirm that the linux amd64 image built?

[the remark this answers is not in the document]
TylerHelmuth: Comparing with the old job, you can see it did the skips also bc it also used a matrix: https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/7832341341/job/21370820540

PR #492, opened 2024-03-04
[chore] Prepare for 0.96.0 release

mx-psi:
Updates open-telemetry/opentelemetry-collector/issues/9681


bryan-aguilar (approved):

distributions/otelcol-contrib/manifest.yaml
```diff
@@ -4,199 +4,199 @@ dist:
   description: OpenTelemetry Collector Contrib
   version: 0.95.1
```
bryan-aguilar, line 5: This was getting removed in a subsequent PR correct? I thought I had seen discussion related to that already.

mx-psi, line 5:
```suggestion
  version: 0.96.0
```

distributions/otelcol/manifest.yaml
```diff
@@ -4,43 +4,43 @@ dist:
   description: OpenTelemetry Collector
   version: 0.95.1
```
mx-psi, line 5:
```suggestion
  version: 0.96.0
```

codeboten (approved):

distributions/otelcol-contrib/manifest.yaml
mx-psi: Thanks for catching this! This is not being removed and it was wrong, I just didn't notice because I replaced `0.95.0` and didn't realize we had made this change

Issue #502, opened 2024-03-19
rpm upgrade of package overwrites config file


KrisBuytaert:
When upgrading the otel-contrib package , it overwrites the existing config file and saves that as config.yaml.rpmsave
As this is clearly a config file,  the preferred behaviour is probably to keep the working config in place and create a /etc/otelcol-contrib/config.yaml.rpmnew  

```
rpm -vUh otelcol-contrib_0.96.0_linux_amd64.rpm
Verifying...                          ################################# [100%]
Preparing...                          ################################# [100%]
Updating / installing...
   1:otelcol-contrib-0:0.96.0-1       warning: /etc/otelcol-contrib/config.yaml saved as /etc/otelcol-contrib/config.yaml.rpmsave

```

PR #503, opened 2024-03-19
[rpm/configfile] change behaviour  when upgrading the package 

KrisBuytaert:
This patch changes the behaviour of how to manage the config file /etc/otelcol-contrib/config.yaml .
It now keeps the existing config file in place and puts the new file as a config.yaml.rpmnew 

```
rpm -vUh otelcol-contrib_0.96.0_linux_amd64.rpm
Verifying... ################################# [100%]
Preparing... ################################# [100%]
Updating / installing...
1:otelcol-contrib-0:0.96.0-1 warning: /etc/otelcol-contrib/config.yaml created as /etc/otelcol-contrib/config.yaml.rpmnew
```

This would fix https://github.com/open-telemetry/opentelemetry-collector-releases/issues/502

### DIFF
--- a/distributions/otelcol-contrib/.goreleaser.yaml
+++ b/distributions/otelcol-contrib/.goreleaser.yaml
@@ -55,7 +55,7 @@ nfpms:
         type: config|noreplace
       - src: config.yaml
         dst: /etc/otelcol-contrib/config.yaml
-        type: config
+        type: config|noreplace
     scripts:
       preinstall: preinstall.sh
       postinstall: postinstall.sh
--- a/distributions/otelcol/.goreleaser.yaml
+++ b/distributions/otelcol/.goreleaser.yaml
@@ -55,7 +55,7 @@ nfpms:
         type: config|noreplace
       - src: config.yaml
         dst: /etc/otelcol/config.yaml
-        type: config
+        type: config|noreplace
     scripts:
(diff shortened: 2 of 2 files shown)


distributions/otelcol-contrib/.goreleaser.yaml
```diff
@@ -55,7 +55,7 @@ nfpms:
         type: config|noreplace
       - src: config.yaml
         dst: /etc/otelcol-contrib/config.yaml
-        type: config
+        type: config|noreplace
```
TylerHelmuth, line 58: Please do the same to the Core distribution

KrisBuytaert: I've also patched the core distribution in the same way :)

TylerHelmuth (commented):
@KrisBuytaert have you been able to test this somehow?

KrisBuytaert:
@TylerHelmuth  yes ,  the above output is the current behaviour with the patch ... before it would have moved the original file aside  like : 

   1:otelcol-contrib-0:0.96.0-1       warning: /etc/otelcol-contrib/config.yaml saved as /etc/otelcol-contrib/config.yaml.rpmsave

jpkrohling (approved):

PR #504, opened 2024-03-21
Automate release notes generation using goreleaser(closes #498)

### LINKED ISSUE #498: Add links to core and contrib to Github releases on releases from this repository
By default, the Github release only includes the changelog from this repository, but most users would care more about the changelog on core and contrib. 

I manually edited the release for v0.96.0 to include the core and contrib changelogs (see https://github.com/open-telemetry/opentelemetry-collector-releases/releases/tag/v0.96.0) but we could make this automatic via goreleaser using https://goreleaser.com/customization/release/#custom-release-notes.

Enzujp:
The .goreleaser file at the root of the project holds default configurations for distro-specific goreleaser files but since we already have configurations tailored to suit the project's needs, I only added a custom formatted string that updates the links to core and contrib with new release versions to be updated and displayed on release, eliminating the need for manual update.


Enzujp:
Hello, @TylerHelmuth Here's what you requested

.gitignore
```diff
@@ -1,6 +0,0 @@
-vendor
```
songy23, line 1: Why delete this file?

Enzujp: It's automatically generated when the .goreleaser file is generated with items not required since we have custom goreleaser files and are only using the default for release-notes generation.

TylerHelmuth: We shouldn't delete our gitignore file

.goreleaser.yaml
```diff
@@ -0,0 +1,3 @@
+release:
+  header: |
+    Check the [v{{.Version}} contrib changelog](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tags/v{{.Version}}) and the [v{{.Version}} core changelog](https://github.com/open-telemetry/opentelemetry-collector/releases/tags/v{{.Version}}) for changelogs on specific components.
```
TylerHelmuth, line 3: Curious, could this be added to each distro's goreleaser file with its specific link?

TylerHelmuth (commented):
Can you link to your test?

.goreleaser.yaml
Enzujp: It can't, not like this anyways. I tried it, the "header" property throws an error or gets ignored on release.
This default .goreleaser only functions when it's at the root of the project folder.

Enzujp: Here's a test repository I created to show how it works, I made a release on there to be sure the formatting was correct. 
Here's a link to it : https://github.com/Enzujp/release-sample

TylerHelmuth: @Enzujp have you tested how this file at the root will interact with our release process which specifies a specific `workdir` that isn't root?

Enzujp: Yes I did, and the goreleaser docs emphasizes the same thing, regardless of the `workdir` the .goreleaser file at the root gets triggered on release and appends the custom formatted header which we've defined in the code.

TylerHelmuth: @Enzujp can you update your test repo to be structured like ours and show that test?

Enzujp: @TylerHelmuth Sure, I used the folder structure and included the .goreleaser files in the `distributions directory`, took out the authentication bits since it's a demo and just ran it as a skeleton to test if it would all work as it should, the failed tests are as a result of the absent bits requiring auth, but it shows that everything runs perfectly fine with the default .goreleaser file at the root of the project, and the header is functional still as you can see from the release. 
Here's the link : https://github.com/Enzujp/sample-releases

TylerHelmuth: Sweet, thank you!

TylerHelmuth (approved):

.goreleaser.yaml
Enzujp: My pleasure! Could you recommend another issue I could work on please ? @TylerHelmuth

.gitgnore
TylerHelmuth, line 1: This file should not be renamed

TylerHelmuth (approved):

songy23 (approved):

jpkrohling (approved):

Issue #383, opened 2023-08-09
Switch to Go 1.21 for official builds


mx-psi:
Same as #286, but for Go 1.21.

Go 1.21 was released recently: https://go.dev/doc/go1.21

After we add testing for it on core (open-telemetry/opentelemetry-collector/issues/8207) and contrib (open-telemetry/opentelemetry-collector-contrib/issues/25116) we can switch to using the Go 1.21 compiler for official builds. Based on historical precedent this will make our life easier when dealing with security issues since we don't depend on the Go team backporting fixes and can react more quickly.

mx-psi:
As mentioned on the Collector SIG, we are working on open-telemetry/opentelemetry-collector/issues/8209 to define the support we give to each platform. I think it would be a good place to discuss and define this.

PR #508, opened 2024-03-22
Enhancement/update dependencies

Enzujp:
Updated Go version to v1.22

### DIFF
--- a/.github/workflows/base-ci-goreleaser.yaml
+++ b/.github/workflows/base-ci-goreleaser.yaml
@@ -45,7 +45,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: '~1.22'
+          go-version: '~1.22.1'
           check-latest: true
 
       - name: Generate the sources
--- a/.github/workflows/base-ci-goreleaser.yaml
+++ b/.github/workflows/base-ci-goreleaser.yaml
@@ -45,7 +45,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: '~1.21.7'
+          go-version: '~1.22'
           check-latest: true
 
       - name: Generate the sources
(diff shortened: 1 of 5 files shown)


.github/workflows/base-ci-goreleaser.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: '~1.21.7'
+          go-version: '~1.22'
```
TylerHelmuth, line 48: We specify the patch version so should be `~1.22.1`

Enzujp: @TylerHelmuth Fixed it now. Good to go?

.gitignore
TylerHelmuth, line 1: Please revert changes to this file

.goreleaser.yaml
```diff
@@ -0,0 +1,3 @@
+release:
```
TylerHelmuth, line 1: This change is unrelated

.gitignore
Enzujp: Hello @TylerHelmuth, please confirm that it is as it should be.

.goreleaser.yaml
Enzujp: @TylerHelmuth I realize that, It's the .goreleaser file - this branch got updated. Should it be taken out ?

TylerHelmuth: Yes

Enzujp: Done

TylerHelmuth (approved):

mx-psi (approved):

jpkrohling (commented):
Shouldn't go.mod be changed as well?

Enzujp:
> Shouldn't go.mod be changed as well?

@jpkrohling I thought about that and then I referenced previous dependency updates #321 and #383 ; go.mod does not need to be manually changed.

PR #522, opened 2024-04-09
remove deprecated component

codeboten:
As per the deprecation message, it was to be no longer supported after the end of 2023

Pinging @evan-bradley for review.


evan-bradley (approved):
Thank you!

PR #525, opened 2024-04-11
Add `org.opencontainers.image.licenses` label to released images

tbrockman:
At `${dayjob}` we currently rely on this OCI label for some automation to determine what level of legal review is required before we're able to use a given image. If one isn't provided, it necessitates some additional red tape. There are definitely ways around it, but I figured the label might be helpful to add for others anyhow.

### DIFF
--- a/cmd/goreleaser/internal/configure.go
+++ b/cmd/goreleaser/internal/configure.go
@@ -193,6 +193,7 @@ func DockerImage(dist, arch, armVersion string) config.Docker {
 			label("revision", ".FullCommit"),
 			label("version", ".Version"),
 			label("source", ".GitURL"),
+			"--label=org.opencontainers.image.licenses=Apache-2.0",
 		},
 		Files:  []string{"config.yaml"},
 		Goos:   "linux",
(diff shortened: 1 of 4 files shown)


TylerHelmuth (approved):

cmd/goreleaser/internal/configure.go
```diff
@@ -193,6 +193,7 @@ func DockerImage(dist, arch, armVersion string) config.Docker {
 			label("revision", ".FullCommit"),
 			label("version", ".Version"),
 			label("source", ".GitURL"),
+			"--label=org.opencontainers.image.licenses=Apache-2.0",
```
swiatekm-sumo, line 196: Why not

```suggestion
			label("org.opencontainers.image.licenses", "Apache-2.0"),
```

tbrockman: Not particularly familiar with goreleaser templating and how it would interpret the resulting:
```yaml
--label=org.opencontainers.image.licenses={{Apache-2.0}}
```

swiatekm-sumo: Ah, I missed that it adds the template braces. In that case, your choice is the correct one.

swiatekm-sumo (approved):

jpkrohling (approved):

PR #526, opened 2024-04-12
use git hash instead of mutable tags and make sure we install latest go

cpanato:
- use git hash instead of mutable tags and make sure we install latest go


jpkrohling (commented):
Will we continue to have dependabot updates for them?

cpanato:
> Will we continue to have dependabot updates for them?

yes, and it will also update the version comments

jpkrohling (approved):

PR #582, opened 2024-06-20
Add OTel-Arrow exporter/receiver to contrib distributions

jmacd:
Adds OTel-Arrow to the contrib distribution.


TylerHelmuth:
We should add it to the k8s distro as well since it is an OTLP receiver/exporter

Issue #295, opened 2023-02-16
release ocb container image


frzifus:
It seems that there has already been some effort to make a ocb container image available. But its unclear where it should live?

- https://github.com/open-telemetry/opentelemetry-collector/issues/5712
- https://github.com/open-telemetry/opentelemetry-collector/pull/6451

According to the discussion in the collector repo, should it be released here?

jpkrohling:
I believe the first step would then to move the release of ocb to this repo. It's currently being released from the core repository: https://github.com/open-telemetry/opentelemetry-collector/releases/tag/cmd%2Fbuilder%2Fv0.72.0

frzifus:
note: As discussed in the collector sig call - we create the image here.

jackgopack4:
Hi all, going to work on adding the ocb release pipeline to the opentelemetry-collector-releases repo and go from there to take this up.

frzifus:
FTR: This topic was discussed during the SIG Call on [March 1, 2023, 17:00 UTC SIG mtg](https://docs.google.com/document/d/1un2i9RmyCJ6J5cs7MoBjwi1GXAN2DLMgu3Z-0kViN70/edit#heading=h.xr6at197y09a). At that time I think recordings have been published on youtube. Unfortunately its not listed on the meeting recording list.

mowies:
Fixed in https://github.com/open-telemetry/opentelemetry-collector-releases/pull/671

PR #608, opened 2024-07-23
Migrate ocb binary release to opentelemetry-collector-releases (see corresponding PR#10710 in opentelemetry-collector repo)

jackgopack4:
**EDIT:** After discussion with the collector SIG call this week, July 31st, I have made small tweaks to the artifact release that will not require an "extra step" of pushing tags. If these changes are accepted, the distributions will be released under tag "vX.XXX.X" and the ocb binaries will be released under "cmd/builder/vX.XXX.X" but both under this repo, opentelemetry-collector-releases. The workflow `builder-release.yaml` will push the additional tags necessary for the builder release.

Refactoring the release process as mentioned in past issues/PRs. ocb binary is currently released in a separate release tag under this repository, which makes it harder to find and inconsistent with the rest of the binaries and images releasing in the opentelemetry-collector-releases repository.

Releasing the ocb binaries in opentelemetry-collector-releases makes this process more standardized and also opens the opportunity to release more tools alongside ocb binary down the road, like corresponding docker images (as discussed in #5712).

#### Link to tracking issue
Fixes [opentelemetry-collector # 10662](https://github.com/open-telemetry/opentelemetry-collector/issues/10662)
Prerequisite for issue #295
[Corresponding PR in opentelemetry-collector](https://github.com/open-telemetry/opentelemetry-collector/pull/10710)

#### Testing
Sample binary release in [forked jackgopack4/opentelemetry-collector-releases repo](https://github.com/jackgopack4/opentelemetry-collector-releases/releases/tag/cmd%2Fbuilder%2Fv0.106.1)

#### Documentation
Updated [corresponding documentation in opentelemetry-collector repository](https://github.com/jackgopack4/opentelemetry-collector/blob/main/docs/release.md) to reflect new release process


.gitignore
jackgopack4, line 1: required in order to pull down core repo to run goreleaser for ocb in the github action workflow without modifying git state (which causes goreleaser to fail)

cmd/builder/.goreleaser.yml
jackgopack4, line 1: migrated from opentelemetry-collector

.github/workflows/builder-testbuild.yaml
```diff
@@ -0,0 +1,61 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@286f3b13b1b49da4ac219696163fb8c1c93e1200 # v6.0.0
+        with:
+          distribution: goreleaser-pro
+          version: latest
+          args: --snapshot --clean -f cmd/builder/.goreleaser.yml
+        env:
+          GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
evan-bradley, line 61:
```suggestion
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}

```
Nit to get rid of the "no newline" character

Makefile
```diff
@@ -40,8 +40,8 @@ ifeq (, $(shell command -v ocb 2>/dev/null))
 	[ "$${machine}" != x86_64 ] || machine=amd64 ;\
 	echo "Installing ocb ($${os}/$${machine}) at $(OTELCOL_BUILDER_DIR)";\
 	mkdir -p $(OTELCOL_BUILDER_DIR) ;\
-	curl -sfLo $(OTELCOL_BUILDER) "https://github.com/open-telemetry/opentelemetry-collector/releases/download/cmd%2Fbuilder%2Fv$(OTELCOL_BUILDER_VERSION)/ocb_$(OTELCOL_BUILDER_VERSION)_$${os}_$${machine}" ;\
-	chmod +x $(OTELCOL_BUILDER) ;\
+	go install go.opentelemetry.io/collector/cmd/builder@v$(OTELCOL_BUILDER_VERSION) ;\
+	mv $$(go env GOPATH)/bin/builder $(OTELCOL_BUILDER) ;\
```
evan-bradley, line 44:
```suggestion
	CGO_ENABLED=0 go build -trimpath -ldflags="-s -w" go.opentelemetry.io/collector/cmd/builder@v$(OTELCOL_BUILDER_VERSION) -o $(OTELCOL_BUILDER) ;\
```
If possible, I wouldn't mind building the module directly:
* Lets us replicate the options used in the goreleaser invocation
* Not a huge deal, but keeps things inside the repo

cmd/builder/.goreleaser.yml
```diff
@@ -0,0 +1,39 @@
...
+  hooks:
+    - go mod download
+monorepo:
+  tag_prefix: cmd/builder/
+  dir: .core/cmd/builder
+builds:
+  - flags:
+      - -trimpath
+    ldflags:
+      - -s -w -X go.opentelemetry.io/collector/cmd/builder/internal.version={{.Version}} -X go.opentelemetry.io/collector/cmd/builder/internal.date={{.Date}}
```
evan-bradley, line 11:
```suggestion
      - -s -w
```
I dug into it, and I don't think we need either of these anymore. `version` is set by OCB itself, and I don't see `date` defined or used anywhere.

.gitignore
```diff
@@ -4,3 +4,4 @@ dist/
 .idea
 .generated-yaml/
 .vscode
+.core
```
evan-bradley, line 7:
```suggestion
.core

```
Another "no newline character" nit

Makefile
jackgopack4: no issues with that approach but I'm going to build and test it in my forked repo for a sanity check

jackgopack4: seems the only way to use go build with a specific version would be to check out the tagged code directly from the opentelemetry-collector repository and then run the build command on the checked out code.

jackgopack4: The following code works locally, keeping the flags etc.:
`CGO_ENABLED=0 go install -trimpath -ldflags="-s -w" go.opentelemetry.io/collector/cmd/builder@v$(OTELCOL_BUILDER_VERSION) ;\`
`mv $$(go env GOPATH)/bin/builder $(OTELCOL_BUILDER) ;\`

cmd/builder/.goreleaser.yml
jackgopack4: committed this change, thanks

.gitignore
jackgopack4: committed this change, thanks

.github/workflows/builder-testbuild.yaml
jackgopack4: committed this change, thanks

Makefile
evan-bradley: Thanks for investigating that. It looks like you're right, `go install` is the right way to go here.

mx-psi (approved):

cmd/builder/.goreleaser.yml
```diff
@@ -0,0 +1,39 @@
...
+    ignore:
+      - goos: windows
+        goarch: arm64
+    binary: ocb
+release:
+  github:
+    owner: open-telemetry
+    name: opentelemetry-collector-releases
+  header: |
+    ### Images and binaries here: https://github.com/open-telemetry/opentelemetry-collector-releases/releases/tag/{{ .Tag }}
```
evan-bradley, line 31: Wouldn't this just point to the current release?

.github/workflows/builder-release.yaml
```diff
@@ -0,0 +1,43 @@
...
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Releases Repo
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          fetch-depth: 0
+      - run: ls -la
+      - name: Push cmd/builder Tag
```
evan-bradley, line 16: Will this have any effect on the release as-is? This workflow runs on the normal `vX.Y.Z` tag and also does the release. My first guess is that GoReleaser will use the tag for the current branch as the release tag.

jackgopack4: [goreleaser lets you specify which tags to use on a build/release job using environment variables](https://goreleaser.com/cookbooks/set-a-custom-git-tag/), specifically referencing the situation in which one commit has multiple tags; this might be something worthwhile to do in order to be safe. I can update the release-contrib, release-core, and release-k8s workflow as well as the builder-release workflow to keep the tags separated via setting the environment variable

cmd/builder/.goreleaser.yml
jackgopack4: yes, it is a remnant from when they were released in the core repo, can probably remove this line

.github/workflows/builder-release.yaml
jackgopack4: of course, if I were to refactor the builder release to join the "partial" build chain that core, contrib, and k8s form under the vX.YYY.Z tag instead of cmd/builder/vX.YYY.Z, this wouldn't be necessary

.github/workflows/builder-release.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Releases Repo
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          fetch-depth: 0
+      - run: ls -la
```
evan-bradley, line 15:
```suggestion
```
This is just debugging code, right?

jackgopack4: yes, thank you. will remove.

.github/workflows/builder-release.yaml
jackgopack4: I've confirmed in my own repo that they do in fact build and release separately, successfully. See here: https://github.com/jackgopack4/opentelemetry-collector-releases/releases

cmd/builder/.goreleaser.yml
jackgopack4: In fact, it points to the main "vX.YYY.Z" release, and would be an easy way for someone to get from the cmd/builder/vX.YYY.Z to the main binaries release, so I think it's ok to leave it in. See link: https://github.com/jackgopack4/opentelemetry-collector-releases/releases/tag/cmd%2Fbuilder%2Fv0.106.1

evan-bradley: I would reword this to say something like "Collector images and binaries here", but that's fine with me.

jackgopack4: "Images and binaries for collector distributions here"

Makefile
```diff
@@ -8,6 +8,9 @@ OTELCOL_BUILDER ?= ${OTELCOL_BUILDER_DIR}/ocb
 DISTRIBUTIONS ?= "otelcol,otelcol-contrib,otelcol-k8s"
 GEN_CONFIG_DISTRIBUTIONS ?= "otelcol,otelcol-contrib"
 
+# required to use specific ocb builder version
+GO111MODULE=on
```
frzifus, line 12: Isnt that deprecated? AFAIK it should no longer has any effect.

frzifus (approved):

Makefile
jackgopack4: hmm, it was failing on the workflow run without it. I can give it a try without in my own repo and see

jackgopack4: yeah not sure why I was getting that error; not happening anymore. I'll remove this variable.

evan-bradley (approved):

dmitryax (approved):

PR #245, opened 2022-11-30
add /bin/sh dependency

shawnbutts:
Fixes #244

PR #617, opened 2024-07-31
Add /bin/sh dependency to linux packages

### LINKED ISSUE #244: RPM packages are missing a /bin/sh dependency
/bin/sh is required for the postinstall.sh and other scripts.

Adding a `Requires:` for all dependencies needed during install is a common standard.

mowies:
Fixes #244 
This re-does the PR already prepared but then closed here: #245


codeboten (approved):
Thanks @mowies

PR #620, opened 2024-08-01
fix linux package dependencies

mowies:
Follow up PR to #244 because the last PR added `/bin/sh` to both the .deb and the .rpm package. But this only works for the RPM package for some reason.
(it also was only requested for the RPM package though)


jackgopack4:
LGTM; were you able to test this in a forked repo or anything?

mowies:
this is actually tested successfully here in this PR directly. see the package tests PR checks :)

mx-psi (approved):

Issue #618, opened 2024-07-31
OTE-01-005 WP1: Linux Binary Hardening Recommendations


mx-psi:
From the [security audit](https://7asecurity.com/reports/pentest-report-opentelemetry.pdf):

>  Testing confirmed that the OpenTelemetry Collector Linux binaries do not leverage a number of compiler flags to mitigate potential memory corruption vulnerabilities, which is a common issue of all Golang-compiled binaries. As a result, the application remains unnecessarily prone to the associated risks.
> Linux binaries fail to leverage the following memory corruption prevention flags:
> - **Stack canaries**: This defense mechanism is used to detect and prevent exploits  from overwriting the return address.
> - **RELRO**: This leaves the GOT section writable. Without the RELRO flag, buffer overflows on a global variable can overwrite GOT entries.
> - **PIE**: The Position Independent Executable (PIE) flag is a security mechanism that enables Address Space Layout Randomization (ASLR), which randomizes the location where system executables are loaded into memory.

Since we do not use CGO, I think this should be as easy as adding `-buildmode=pie`.

> [!NOTE]  
> [2024 OpenTelemetry security audit](https://opentelemetry.io/blog/2024/security-audit-results/) finding reference:
> **OTE-01-005 WP1**: Linux Binary Hardening Recommendations

jackgopack4:
seems like buildmode=pie doesn't necessarily do anything if CGO is disabled

mx-psi:
@jackgopack4 I believe some platforms do support it and others do not (at least that's what I get from this comment: https://github.com/golang/go/issues/64875#issuecomment-1874249860 ). We could pass the option only in those platforms

jackgopack4:
@mx-psi you're right; the error message is exactly the same and only happens on unsupported platforms. will add conditional logic and test again.

jackgopack4:
apologies for the confusion

PR #626, opened 2024-08-02
enabled CGO for Macos otelcol contrib

pureklkl:
Add a CGO enabled otelcol-contrib for macos
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33393

There are some test failures need be fixed by this PR
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33921


atoulme:
Nice work :)

jpkrohling (changes requested):
Can you please go over our security audit? it lists a few recommendations in case we ever change to CGO, which this PR is doing.

https://7asecurity.com/reports/pentest-report-opentelemetry.pdf

jpkrohling:
Related: https://github.com/open-telemetry/opentelemetry-collector-releases/issues/618

pureklkl:
The checksec.sh is not working for macos [ref](https://github.com/slimm609/checksec.sh?tab=readme-ov-file#for-osx)
I used [snake&apple](https://github.com/Karmaz95/Snake_Apple/tree/main?tab=readme-ov-file#snake--apple) instead, see section3. But this is also not accurate because it is for Xcode compiled binary instead of golang. Here is the result

PIE:            True
ARC:            False
STRIPPED:       False
CANARY:         True
NX STACK:       True
NX HEAP:        False
XN:             True
NOTARIZED:      False
ENCRYPTED:      False
RESTRICTED:     False
HARDENED:       False
APP SANDBOX:    False
FORTIFIED:      True
RPATH:          False

I am not a security expert, so need advice other tools that can check macos binary security.

.github/workflows/release-contrib-cgo.yaml
```diff
@@ -0,0 +1,17 @@
+name: Release Contrib
```
jpkrohling, line 1:
```suggestion
name: Release Contrib with CGO
```

distributions/otelcol-contrib-cgo/README.md
```diff
@@ -0,0 +1,11 @@
+# OpenTelemetry Collector Contrib CGO Distro
```
jpkrohling, line 1: As a user, I would find it confusing. Why should I care? What does it bring that others don't? 

I'm not sure I like the idea of a distribution with just a different platform or compilation mechanism. I would prefer a solution where we built with the right options based on what we need to ship.

Does hostmetrics require cgo when compiled for Darwin? Then we should use cgo to compile the distributions that include that component for Darwin.

I find a separate cgo distribution confusing, and opens the door for a cgo distribution for other platforms, which is not something we want.

jpkrohling (changes requested):

distributions/otelcol-contrib-cgo/README.md
pureklkl: My concern is that when CGO is enabled, the binary is coupled to certain OS versions and will support less OS versions.

pureklkl: @jpkrohling 
Hi, I have updated the PR to just enable CGO for Macos otelcol contrib, could you review it again?

.github/workflows/base-release.yaml
```diff
@@ -51,8 +51,12 @@ jobs:
         with:
           platforms: arm64,ppc64le,linux/arm/v7,s390x
 
-      - uses: docker/setup-buildx-action@aa33708b10e362ff993539393ff100fa93ed6a27 # v3.5.0
+<<<<<<< HEAD
```
jpkrohling, line 54: This doesn't look right :-)

jpkrohling (changes requested):

distributions/otelcol-contrib-cgo/README.md
jpkrohling: I would like to have it discussed as part of the SIG and have buy-in from other maintainers.

pureklkl:
configure.go need to be modified to fix the build, and it will be affected by whether we want a new distribution. I will fix the build once the community have a decision.

distributions/otelcol-contrib-cgo/README.md
pureklkl: Hi, I have tested with golang 10.15 and it works, which is also the oldest macos version that golang >= 1.21 supported [ref](https://go.dev/wiki/MinimumRequirements)
I think we could just turn on the CGO option without a separated distribution.

cmd/goreleaser/internal/configure.go
```diff
@@ -90,19 +98,58 @@ func Build(dist string) config.Build {
...
+		Archive(dist, cgo),
 	}
 }
 
 // Archive configures a goreleaser archive (tarball).
 // https://goreleaser.com/customization/archive/
-func Archive(dist string) config.Archive {
+func Archive(dist string, cgo string) config.Archive {
+	builds := []string{dist}
+	if len(cgo) > 0 {
```
jackgopack4, line 146: what's the benefit of having cgo set as a string instead of a bool? Do we plan to have more conditional logic based on the content of `cgo` if we have more distributions built with cgo in the future?

jackgopack4:
have you made a modified fork of this pr with references to repos changed to your fork, and run a release in your fork?

cmd/goreleaser/internal/configure.go
pureklkl: I put it to string in case we need other OS to support cgo in the future. I also wrote it in the cgoFlag description.

pureklkl:
https://github.com/shirou/gopsutil/releases/tag/v4.24.9

gopsutil removed the cgo requirement, no action needed.

PR #630, opened 2024-08-07
[chore] remove APK package releases

### LINKED ISSUE #616: Why is there an APK package that contains a systemd service (non-compatible with Alpine)?
What is the use case of the APK package releases containing a systemd service? From my understanding, Alpine Linux is not compatible with systemd services and would need a different service setup to work correctly using open-rc.

When installing the package on Alpine, I don't get any errors, files are copied correctly into the right places, but also the service is not started, so I guess it fails somewhat silently.

mowies:
### This PR
- removes the releases for Alpine Linux packages (.apk) since they are currently not compatible with Alpine as discussed in the collector SIG meeting on August 7th
- fixes #616

As a follow-up and if there is somebody to take care of it, [this is the issue](https://github.com/open-telemetry/opentelemetry-collector-releases/issues/631) to properly implement an Alpine compatible linux package of the collector.


jackgopack4:
should we update the documentation (README in both base folder and distributions folder) to remove reference to apk versions?

mowies:
go. catch @jackgopack4 , done :)

jpkrohling (approved):

PR #638, opened 2024-08-14
Bump to Go 1.23 for all builds

mx-psi:
Since we have open-telemetry/opentelemetry-collector-contrib/pull/34658 and open-telemetry/opentelemetry-collector/pull/10869 we can do the same as in #471

PR #640, opened 2024-08-14
Fix goreleaser ci

jackgopack4:
Fixes issue raised in https://github.com/open-telemetry/opentelemetry-collector-releases/pull/638.

Adds monorepo tag to goreleaser files in order to specify that "vX.YYY.Z" tags are the only one to pay attention to for main release ("cmd/builder/v..." tags will only apply for ocb release)


mx-psi (approved):
Thank you!

Issue #657, opened 2024-09-04
Create a v1 otlp only distro


codeboten:
As part of https://github.com/open-telemetry/opentelemetry-collector/issues/9375, we'll need a distro for this

TylerHelmuth:
As a formality, here is how such a distribution meets our criteria:

1. **Serve a specific purpose that has minimal overlap with the purpose of any other distribution**: This distribution's specific purpose would be to be used as a "side-car" - an absolutely bare minimum collector capable of receiving application OTLP telemetry, queuing, and exporting that OTLP telemetry.
2. **Should meet general needs and be desired by many users.**: A side car collector has been brought up before and I believe implemented elsewhere (@jpkrohling can you confirm?). It could be usable as the default side-car collector for the OpenTelemetry Operator.
3. **Should not be specific to any vendor.**: ✅ 
4. **May be focused on development or proof of concept use cases.  The distribution should clearly indicate whether the Collector SIG recommends the distribution be used in production environments.**: This will be a production-ready distribution.
5. **Must only include components from the `opentelemetry-collector` and `opentelemetry-collector-contrib` repositories**: ✅ 
6. **Have a clearly defined list of criteria for which components are included.**: The criteria will be only only the `otlpreceiver`, `otlpexporter`, and `otlphttpexporter` are allowed.
7. **Must include the following assets except where the specific purpose of the distribution is naturally associated with a subset of these assets.**: We'll release our typical assets, including windows.

jpkrohling:
> A side car collector has been brought up before and I believe implemented elsewhere (@jpkrohling can you confirm?).

Yes, here: https://github.com/jpkrohling/otelcol-distributions/tree/main/distributions/otelcol-otlp

That one includes a few more things that I thought were essential after putting this in the real world. Here's the full manifest:

https://github.com/jpkrohling/otelcol-distributions/blob/5566bf802a2177effb98894f47aab6ab6a1b9696/distributions/otelcol-otlp/manifest.yaml#L9-L21

jade-guiton-dd:
> @jade-guiton-dd would you like to take this one?

Sure.

mx-psi:
Fixed by #663

PR #507, opened 2024-03-21
Add a Kubernetes Distribution

TylerHelmuth:
This PR adds a new distribution specifically for monitoring Kubernetes. For full details see https://github.com/open-telemetry/opentelemetry-collector-releases/issues/357.

Tested in https://github.com/TylerHelmuth/opentelemetry-collector-releases:
  - action: https://github.com/TylerHelmuth/opentelemetry-collector-releases/actions/runs/8470931109
  - release: https://github.com/TylerHelmuth/opentelemetry-collector-releases/releases/tag/v0.97.7
  - packages: 
    - https://github.com/TylerHelmuth/opentelemetry-collector-releases/pkgs/container/opentelemetry-collector-releases%2Fopentelemetry-collector-k8s
    - https://hub.docker.com/r/thelmuth34/opentelemetry-collector-k8s/tags

Example helm chart values.yaml using this distro:

```yaml
mode: daemonset

clusterRole:
  create: true
  rules:
    - apiGroups: 
        - ""
      resources:
        - nodes/proxy
      verbs:
        - get

resources:
  limits:
    cpu: 200m
    memory: 512Mi
  requests:
    cpu: 100m
    memory: 256Mi

image:
  repository: thelmuth34/opentelemetry-collector-k8s
  tag: 0.97.7

command:
  name: otelcol-k8s

presets:
  logsCollection:
    enabled: true
  kubernetesAttributes:
    enabled: true
    extractAllPodLabels: true
    extractAllPodAnnotations: true
  kubeletMetrics:
    enabled: true

config:
  receivers:
    jaeger: null
    zipkin: null
    kubeletstats:
      collection_interval: 30s
      metric_groups:
        - node
        - pod
      metrics:
        k8s.node.uptime:
          enabled: true
        k8s.pod.uptime:
          enabled: true
        k8s.pod.cpu_limit_utilization:
          enabled: true
        k8s.pod.cpu_request_utilization:
          enabled: true
        k8s.pod.memory_limit_utilization:
          enabled: true
        k8s.pod.memory_request_utilization:
          enabled: true

  service:
    pipelines:
      traces:
        receivers: [otlp]
        exporters: [debug]
      metrics:
        exporters: [debug]
      logs:
        exporters: [debug]

ports:
  jaeger-compact:
    enabled: false
  jaeger-thrift:
    enabled: false
  jaeger-grpc:
    enabled: false
  zipkin:
    enabled: false
```

```yaml
mode: deployment

image:
[...]

PR #663, opened 2024-09-18
Add minimal OTLP-only distribution

jade-guiton-dd:
As per #657, this PR adds a new distribution, tentatively named `otelcol-minimal`, which only contains the OTLP receivers and exporters (both for gRPC and HTTP), as well as the file and env config providers (the two that have v1).


distributions/otelcol-minimal/manifest.yaml
```diff
@@ -0,0 +1,18 @@
+dist:
+  module: github.com/open-telemetry/opentelemetry-collector-releases/minimal
+  name: otelcol-minimal
+  description: OpenTelemetry Collector Minimal
+  version: 0.109.0
+  output_path: ./_build
+  otelcol_version: 0.109.0
+
+exporters:
```
jpkrohling, line 9: I had a similar distribution myself and it was very similar to this one at first, but found that I was missing a couple of things:

- at least one auth extension
- pprof extension
- debug exporter

Without those, I think the Collector cannot be properly monitored, secured, and configured (debug exporter is useful when setting up pipelines in general).

I also ended up adding this, but I think we should have v1 without it, adding it just after we have the healthcheck v2 in place:

- healthcheckextension

.github/workflows/ci-goreleaser-minimal.yaml
```diff
@@ -0,0 +1,33 @@
+name: Continuous Integration - Minimal - GoReleaser
```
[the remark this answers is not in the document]
mx-psi: We discussed this on the 2024-09-18 Collector stability meeting, the general sentiment was that OTLP is the right call here

distributions/otelcol-minimal/manifest.yaml
mx-psi: We discussed this on the 2024-09-18 Collector stability meeting; we can start with just otlpreceiver, otlpexporter and otlphttpexporter for now and open dedicated issues to add the other components mentioned here at some point (which may be after 1.0 and instead in a later 1.x release). Once we merge this PR I will file issues to add these components to the distro.

.github/workflows/base-ci-goreleaser.yaml
```diff
@@ -67,7 +67,7 @@ jobs:
         uses: goreleaser/goreleaser-action@286f3b13b1b49da4ac219696163fb8c1c93e1200 # v6.0.0
         with:
           distribution: goreleaser-pro
-          version: latest
+          version: v2.2.0
```
mx-psi, line 70: Could you file a separate PR for this?

distributions/otelcol-otlp/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+Unlike the Core and Contrib distributions, the deb/rpm/msi installers and the Docker images for this distribution do not provide a default configuration file. One will need to be created in the appropriate location (`/etc/otelcol-otlp/config.yaml` or `%ProgramW6432%\OpenTelemetry Collector\config.yaml` by default).
+
+## Components
+
+The full list of components is available in the [manifest](manifest.yaml)
+
+### Rules for Component Inclusion
+
+- Only `otlpreceiver`, `otlpexporter`, and `otlphttpexporter` are allowed.
```
mx-psi, line 15: After the discussion from yesterday, I feel like we should leave this more open. We will start with these three components, but may add more to the list as time goes by

@TylerHelmuth @jpkrohling @codeboten Any suggestions on the wording here?

distributions/otelcol-otlp/otelcol-otlp.conf
```diff
@@ -0,0 +1,5 @@
+# Systemd environment file for the otelcol-otlp service
+
+# Command-line options for the otelcol-otlp service.
+# Run `/usr/bin/otelcol-otlp --help` to see all available options.
```
mx-psi, line 4: Should we add a comment here telling the user to create the configuration file in the `/etc/otelcol-otlp/config.yaml` file?

.github/workflows/base-ci-goreleaser.yaml
jade-guiton-dd: I filed it as #667. Should I remove the change from this PR?

mx-psi: It should go away after I merge #667 and you merge main into this PR

jpkrohling (approved):
Approving, so that we can get it out for feedback.

distributions/otelcol-otlp/README.md
jpkrohling: I think we can adapt this based on the feedback we receive from users.

mx-psi (approved):

Makefile
```diff
@@ -5,7 +5,7 @@ OTELCOL_BUILDER_VERSION ?= 0.109.0
 OTELCOL_BUILDER_DIR ?= ${HOME}/bin
 OTELCOL_BUILDER ?= ${OTELCOL_BUILDER_DIR}/ocb
 
-DISTRIBUTIONS ?= "otelcol,otelcol-contrib,otelcol-k8s"
+DISTRIBUTIONS ?= "otelcol,otelcol-contrib,otelcol-k8s,otelcol-otlp"
 GEN_CONFIG_DISTRIBUTIONS ?= "otelcol,otelcol-contrib"
```
TylerHelmuth, line 9: This distribution's goreleaser file should be generated. Unlike the k8s distro that has unique artifact requirements, this distribution's artifacts should match contrib and core.

distributions/otelcol-otlp/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+FROM scratch
+
+ARG USER_UID=10001
+USER ${USER_UID}
+
+COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --chmod=755 otelcol-otlp /otelcol-otlp
+ENTRYPOINT ["/otelcol-otlp"]
+CMD ["--config", "/etc/otelcol-otlp/config.yaml"]
```
TylerHelmuth, line 12: For the k8s distribution we [stopped including a default config](https://github.com/open-telemetry/opentelemetry-collector-releases/pull/507#discussion_r1541420891).  I believe the some logic applies here.

distributions/otelcol-otlp/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+EXPOSE 4317 55678 55679
```
TylerHelmuth, line 13: Do `55678 55679` need exposed? I would think only `4317 4318` is needed.

TylerHelmuth (changes requested):

Makefile
jade-guiton-dd: The problem is that the generated goreleaser file expects a config file to install. If we want to stop including a default config, we would have to make the generator much more flexible. Should I work on that in this PR?

distributions/otelcol-otlp/Dockerfile
jade-guiton-dd: I didn't include a default config file, but I kept the default location, as I thought it wouldn't have those problems, and it would avoid an extra step during step. Should I remove it?

distributions/otelcol-otlp/Dockerfile
jade-guiton-dd: You're right, fixed.

distributions/otelcol-otlp/Dockerfile
mx-psi: I feel like adding this default path in the systemd service definition makes sense, but here we can just remove the `CMD` part and force users to provide their own command line arguments

Makefile
mx-psi: If we need to do this, I vote for doing the refactor in a separate PR that we merge before this one

TylerHelmuth: Ok, we'll move forward without generating the goreleaser file.

songy23 (approved):

.github/workflows/release-otlp.yaml
```diff
@@ -0,0 +1,16 @@
+name: Release OTLP
```
TylerHelmuth, line 1: @jade-guiton-dd were you able to test this release in a Fork? We've had a lot of failed releases in the past because the job wasn't tested so now we require it.

jade-guiton-dd: I haven't tested it yet. Is there something I need to setup on my fork for the CI to work, or do I just need to create a "vX" tag?

TylerHelmuth: You'll need a goreleaser-pro key.  @mx-psi or @jpkrohling can help if you don't already have access to one.

jade-guiton-dd: I tested the release CI on my fork, and after a few adjustments (4c1f276) to point to my own Docker repo, and working around a problem which was currently affecting all release workflows (#676), it seems to work ([Github release](https://github.com/jade-guiton-dd/opentelemetry-collector-releases/releases/tag/v0.109.0-test3), [Docker repo](https://hub.docker.com/repository/docker/jadeguiton994/otelcol-release-tests/tags) if you want to check the result).

TylerHelmuth (approved):

PR #667, opened 2024-09-19
Pin CI goreleaser version to v2.2.0

jade-guiton-dd:
Version 2.3 of goreleaser changed some of the output directory names, making the `Upload linux service packages` task fail silently, causing the `Package tests` tasks to fail at retrieving the package artifacts under test.

This PR pins the version of goreleaser used in the CI to the previous one, v2.2.0, to resolve the issue.

We could also pin to v2.3 and update the expected directory name, but it seems that the name of output directories are not considered stable by goreleaser, so I think a better long-term solution is needed.


mx-psi (approved):

mx-psi:
I can merge this once it is marked as ready for review :)

mowies (approved):

PR #678, opened 2024-09-25
[chore] refactor(contrib): remove replace directive for `go-ieproxy` package

andrzej-stencel:
The replacement for the package `github.com/mattn/go-ieproxy` is currently in effect and replaces `v0.0.11` of the package with version `v0.0.1`. I believe this replacement is not necessary anymore, as the issue https://github.com/mattn/go-ieproxy/issues/45 was fixed in https://github.com/mattn/go-ieproxy/pull/47 in v0.0.11.

If I understand the original [issue](https://github.com/mattn/go-ieproxy/issues/45) correctly, it being fixed can be verified by a successful Windows build in GitHub Actions. Perhaps @codeboten you can take a look at this and possibly correct me, as you originally [introduced](https://github.com/open-telemetry/opentelemetry-collector-releases/pull/242) this replace directive.


codeboten (approved):
Removed it in contrib and confirmed that all checks were still passing 👍🏻

songy23 (approved):

PR #684, opened 2024-10-02
[chore] add update version automation workflow for artifacts release

jackgopack4:
Add quality-of-life enhancement to automate bumping version number in the Makefile as well as the various distribution manifests (as the number of distros grows, this step will become more of a pain).

Feel free to run in your own fork to verify; also see the successful run in my fork: 
https://github.com/jackgopack4/opentelemetry-collector-releases/actions/runs/12320554731/job/34390067049
As well as the generated PR in my fork:
https://github.com/jackgopack4/opentelemetry-collector-releases/pull/23

If this gets merged I will open a PR to update collector release doc in `opentelemetry-collector` repo.

Automatic version bump (increase minor version number) is default behavior; this can be overridden by providing a manual version number update (like bugfix v0.110.0 -> v0.110.1)

### DIFF
--- a/.github/workflows/scripts/bump-versions.sh
+++ b/.github/workflows/scripts/bump-versions.sh
@@ -1,10 +1,9 @@
 #!/bin/bash
 set -e
-# This script takes either current beta version, current stable version, or both,
-# and optionally next beta version and next stable version, and updates the version
-# in the specified files. If next version is not provided, it will infer the next 
-# semantic version (e.g. v0.110.0 -> v0.111.0 or v1.16.0 -> v1.17.0) based on the 
-# current version(s) passed.
+# This script reads current versions and takes optional next versions, and updates the
+# version in the specified files. If next version is not provided, it will infer the 
+# next semantic version (e.g. v0.110.0 -> v0.111.0 or v1.16.0 -> v1.17.0) based on the 
+# current version(s) read in.
 
 # List of files to update
 manifest_files=(
@@ -154,6 +153,10 @@ else
 fi
 
 # Update versions in each manifest file
+echo "Updating core beta version from $current_beta_core to $next_beta_core,"
+echo "core stable version from $current_stable to $next_stable,"
+echo "contrib beta version from $current_beta_contrib to $next_beta_contrib,"
+echo "and distribution version to $next_distribution_version"
 for file in "${manifest_files[@]}"; do
   if [ -f "$file" ]; then
     $sed_command "s/\(^.*go\.opentelemetry\.io\/collector\/.*\) v$escaped_current_beta_core/\1 v$next_beta_core/" "$file"
@@ -174,7 +177,7 @@ echo "Version update completed."
 make chlog-update VERSION="v$next_distribution_version"
 
 # Commit changes and draft PR
-if [ "$commit_changes" = false ] && [ "$create_pr" = false ]; then
+if [ "$commit_changes" = false ]; then
   echo "Changes not committed and PR not created."
   exit 0
 fi
@@ -230,4 +233,5 @@ else
   fi
 fi
 
-echo "Changes committed and PR created."
+echo "Changes committed and PR created:"
+gh pr view
--- a/.github/workflows/scripts/bump-versions.sh
+++ b/.github/workflows/scripts/bump-versions.sh
@@ -1,0 +1,233 @@
+#!/bin/bash
+set -e
+# This script takes either current beta version, current stable version, or both,
+# and optionally next beta version and next stable version, and updates the version
+# in the specified files. If next version is not provided, it will infer the next 
+# semantic version (e.g. v0.110.0 -> v0.111.0 or v1.16.0 -> v1.17.0) based on the 
+# current version(s) passed.
+
+# List of files to update
+manifest_files=(
+  "distributions/otelcol-contrib/manifest.yaml"
+  "distributions/otelcol/manifest.yaml"
+  "distributions/otelcol-k8s/manifest.yaml"
+  "distributions/otelcol-otlp/manifest.yaml"
+)
+
+# Function to display usage
+usage() {
+  echo "Usage: $0 [--commit] [--pull-request] [--next-beta-core <next-beta-core>] [--next-beta-contrib <next-beta-contrib>] [--next-stable <next-stable>]"
+  echo "  --next-beta-core: Next beta version of the core component (e.g., v0.111.0)"
+  echo "  --next-beta-contrib: Next beta version of the contrib component (e.g., v0.111.0)"
+  echo "  --next-stable: Next stable version of the core component (e.g., v1.17.0)"
+  echo
+  echo "  --commit: Commit the changes to a new branch"
+  echo "  --pull-request: Push the changes to the repo and create a draft PR (requires --commit)"
+  exit 1
+}
+
+# Function to validate semantic version and strip leading 'v'
+validate_and_strip_version() {
+  local var_name=$1
+  local version=${!var_name}
+  # Strip leading 'v' if present
+  version=${version#v}
+  if [[ ! $version =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+    echo "Invalid version: $version. Must be a semantic version (e.g., 1.2.3)."
+    exit 1
+  fi
+  eval "$var_name='$version'"
+}
+commit_changes=false
+create_pr=false
+# Parse named arguments
+current_beta_core=$(awk '/^.*go\.opentelemetry\.io\/collector\/.* v0/ {print $4; exit}' distributions/otelcol/manifest.yaml)
+current_beta_contrib=$(awk '/^.*github\.com\/open-telemetry\/opentelemetry-collector-contrib\/.* v0/ {print $4; exit}' distributions/otelcol-contrib/manifest.yaml)
+current_stable=$(awk '/^.*go\.opentelemetry\.io\/collector\/.* v1/ {print $4; exit}' distributions/otelcol/manifest.yaml)
+while [[ "$#" -gt 0 ]]; do
+  case $1 in
+    --next-beta-core) next_beta_core="$2"; shift ;;
+    --next-beta-contrib) next_beta_contrib="$2"; shift ;;
+    --next-stable) next_stable="$2"; shift ;;
+    --commit) commit_changes=true ;;
+    --pull-request) create_pr=true ;;
+    *) echo "Unknown parameter passed: $1"; usage ;;
+  esac
+  shift
+done
+
+# Check if --pull-request is passed without --commit
+if [ "$create_pr" = true ] && [ "$commit_changes" = false ]; then
+  echo "--pull-request requires --commit"
+  usage
+fi
+
+# Validate and strip versions
+if [ -n "$current_beta_core" ]; then
+  validate_and_strip_version current_beta_core
+fi
+if [ -n "$current_beta_contrib" ]; then
+  validate_and_strip_version current_beta_contrib
+fi
+if [ -n "$current_stable" ]; then
+  validate_and_strip_version current_stable
+fi
+if [ -n "$next_beta_core" ]; then
+  validate_and_strip_version next_beta_core
+fi
+if [ -n "$next_beta_contrib" ]; then
+  validate_and_strip_version next_beta_contrib
+fi
+if [ -n "$next_stable" ]; then
+  validate_and_strip_version next_stable
+fi
+
+# Function to compare two semantic versions and return the maximum
+max_version() {
+  local version1=$1
+  local version2=$2
+
+  # Strip leading 'v' if present
+  version1=${version1#v}
+  version2=${version2#v}
+
+  # Split versions into components
+  IFS='.' read -r -a ver1 <<< "$version1"
+  IFS='.' read -r -a ver2 <<< "$version2"
+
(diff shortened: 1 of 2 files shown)


andrzej-stencel (commented):
I :100: agree that bumping the version should be a simple, single step. This change implements this as a series of steps in GitHub Actions, which makes it not usable if you want to run it outside of GitHub Actions. I'd rather see a shell script written that does the version updates, and that script called as a single step in GitHub Actions. Committing changes should be separate from the script, so that anyone can run the script locally and preview the changes before making it a commit.

Note that there are (to my knowledge) three potentially different versions that need to be bumped in the manifests, as described [here](https://github.com/open-telemetry/opentelemetry-collector/issues/11129#issuecomment-2343499207):

> - stable components all use the stable version - currently v1.15.0
> - unstable components from core repository use the core's version - currently v0.109.0
> - unstable components from contrib repository use the contrib's version - currently v0.109.0
> 
> The core and contrib versions are usually the same, but they are not guaranteed to be the same.

jackgopack4:
> I 💯 agree that bumping the version should be a simple, single step. This change implements this as a series of steps in GitHub Actions, which makes it not usable if you want to run it outside of GitHub Actions. I'd rather see a shell script written that does the version updates, and that script called as a single step in GitHub Actions. Committing changes should be separate from the script, so that anyone can run the script locally and preview the changes before making it a commit.

Would be happy to do it this way, just was modeling it after the existing action script in collector core: https://github.com/open-telemetry/opentelemetry-collector/blob/main/.github/workflows/prepare-release.yml

> Note that there are (to my knowledge) three potentially different versions that need to be bumped in the manifests, as described [here](https://github.com/open-telemetry/opentelemetry-collector/issues/11129#issuecomment-2343499207):
> 
> > * stable components all use the stable version - currently v1.15.0
> > * unstable components from core repository use the core's version - currently v0.109.0
> > * unstable components from contrib repository use the contrib's version - currently v0.109.0

Got it, will add stable and beta component update regardless of the method of execution

jackgopack4:
@andrzej-stencel made the changes you requested; runs in one bash script that can be ran locally in addition to in github actions and adds stable/beta functionality

.github/workflows/scripts/bump-versions.sh
```diff
@@ -0,0 +1,144 @@
...
+    fi
+    rm "${file}.bak"
+  else
+    echo "File $file does not exist"
+  fi
+done
+
+echo "Version update completed."
+
+Commit changes and draft PR
```
andrzej-stencel, line 112: As mentioned in https://github.com/open-telemetry/opentelemetry-collector-releases/pull/684#pullrequestreview-2345254646:

> Committing changes should be separate from the script, so that anyone can run the script locally and preview the changes before making it a commit.

At the very least, the commit, push, PR creation part should be opt-in via a script option like `--commit` or similar. The default (when the `--commit` option is not passed to the script) should be to not make the commit.

.github/workflows/update-version.yaml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/checkout@v2
+
+      - name: Run bump-versions.sh
+        run: |
+          chmod +x .github/workflows/scripts/bump-versions.sh
+          .github/workflows/scripts/bump-versions.sh \
+            --current_beta "${{ github.event.inputs.current_beta }}" \
+            --current_stable "${{ github.event.inputs.current_stable }}" \
+            --next_beta "${{ github.event.inputs.next_beta }}" \
+            --next_stable "${{ github.event.inputs.next_stable }}"
```
andrzej-stencel, line 45: As mentioned in https://github.com/open-telemetry/opentelemetry-collector-releases/pull/684#pullrequestreview-2345254646, there are three distinct versions used in the builder configurations. Two of those are equal _most of the time_. This script assumes that core and contrib versions are always the same, which means it will do the wrong thing in case there's a mismatch between core and contrib versions.

I'm reluctant to have an automation in place that works correctly _most of the time_, as opposed to _always_. Releasing software should be as frictionless as possible, given that it is sometimes done under pressure (e.g. for hot fix releases).

I propose to update the script to accept three versions, and perform the replacements separately for core and contrib components. It is possible to easily tell those apart as the module paths are different.

.github/workflows/scripts/bump-versions.sh
```diff
@@ -0,0 +1,144 @@
...
+
+# List of files to update
+# TODO: Uncomment cmd/builder/builder-config.yaml once PR #671 merged
+files=(
+  "distributions/otelcol-contrib/manifest.yaml"
+  "distributions/otelcol/manifest.yaml"
+  "distributions/otelcol-k8s/manifest.yaml"
+  "distributions/otelcol-otlp/manifest.yaml"
+  "Makefile"
+  # "cmd/builder/builder-config-yaml"
```
andrzej-stencel, line 90: This can be removed, as https://github.com/open-telemetry/opentelemetry-collector-releases/pull/671/ does not add this file anymore.

.github/workflows/scripts/bump-versions.sh
```diff
@@ -0,0 +1,144 @@
...
+
+# This script takes either current beta version, current stable version, or both,
+# and optionally next beta version and next stable version, and updates the version
+# in the specified files. If next version is not provided, it will infer the next 
+# semantic version (e.g. v0.110.0 -> v0.111.0 or v1.16.0 -> v1.17.0) based on the 
+# current version(s) passed.
+
+# Function to display usage
+usage() {
+  echo "Usage: $0 --current_beta <current_beta> [--current_stable <current_stable>] [--next_beta <next_beta>] [--next_stable <next_stable>]"
```
andrzej-stencel, line 11: This is a nit, but words in option names should probably be separated with a dash `-` and not an underscore `_`.
```suggestion
  echo "Usage: $0 --current-beta <current_beta> [--current-stable <current_stable>] [--next-beta <next_beta>] [--next-stable <next_stable>]"
```

.github/workflows/scripts/bump-versions.sh
```diff
@@ -0,0 +1,144 @@
...
+  "distributions/otelcol-otlp/manifest.yaml"
+  "Makefile"
+  # "cmd/builder/builder-config-yaml"
+)
+
+# Update versions in each file
+for file in "${files[@]}"; do
+  if [ -f "$file" ]; then
+    if [ -n "$current_beta" ]; then
+      sed -i.bak "s/$current_beta/$next_beta/g" "$file"
```
andrzej-stencel, line 97: We don't need the `g` option, right? The version only appears once on each line.
```suggestion
      sed -i.bak "s/$current_beta/$next_beta/" "$file"
```

.github/workflows/update-version.yaml
jackgopack4: apologies, I must not have read your comment thoroughly enough. I agree with your proposed change and will push an update to the PR branch shortly.

.github/workflows/scripts/bump-versions.sh
jackgopack4: fixed, thanks

.github/workflows/scripts/bump-versions.sh
jackgopack4: fixed, thanks

.github/workflows/scripts/bump-versions.sh
jackgopack4: added two flags, --commit which commits changes and creates new branch, pushing to remote, and --pull-request which creates draft PR with that new branch. Github workflow passes --commit every time, but allows release manager to check or uncheck --pull-request (checked by default).

.github/workflows/scripts/bump-versions.sh
```diff
@@ -0,0 +1,210 @@
...
+          sed -i.bak "${line_number}s/$current_beta_core/$next_beta_core/" "$temp_file"
+        fi
+        if [ -n "$current_stable" ]; then
+          sed -i.bak "${line_number}s/$current_stable/$next_stable/" "$temp_file"
+        fi
+      fi
+      line_number=$((line_number + 1))
+    done < "$file"
+    mv "$temp_file" "$file"
+    rm "${temp_file}.bak"
```
andrzej-stencel, line 143: The script logs errors trying to remove non-existent `.bak` files when run with only the `--current-beta-contrib` flag:

```console
./.github/workflows/scripts/bump-versions.sh --current-beta-contrib v0.111.0
rm: cannot remove '/tmp/tmp.h6AQVOKLTG.bak': No such file or directory
rm: cannot remove '/tmp/tmp.VsQcTga32N.bak': No such file or directory
Version update completed.
Changes not committed and PR not created.
```

I propose to add an existence check before removing the `.bak` files.

```suggestion
    [ -f "${temp_file}.bak" ] && rm "${temp_file}.bak"
```

.github/workflows/scripts/bump-versions.sh
```diff
@@ -0,0 +1,210 @@
...
+  "Makefile"
+)
+
+# Update versions in each file
+for file in "${files[@]}"; do
+  if [ -f "$file" ]; then
+    temp_file=$(mktemp)
+    cp "$file" "$temp_file"
+    line_number=1
+    while IFS= read -r line; do
```
andrzej-stencel, line 127: Why does the script run `sed` for each line separately? Putting efficiency aside, this seems harder to understand than just running `sed` on the whole file(s). The replacement patterns are also less precise. For example, there are lines in `Makefile` that have `github.com/open-telemetry/opentelemetry-collector-contrib` in them, but are not subject to replacements.

I would imagine running the `sed` command on the whole files and specifying the replacement patterns specifically for each purpose. For example:
- to replace core versions in manifests: `sed "s/\(^.*go\.opentelemetry\.io\/collector\/.*\) v0\.111\.0/\1 v0.112.0/" distributions/otelcol/manifest.yaml`
- to replace `otelcol_version` in manifests: `sed "s/\(^\s\+otelcol_version:\) 0\.111\.0/\1 0.112.0/" distributions/otelcol/manifest.yaml`
- etc.

.github/workflows/scripts/bump-versions.sh
```diff
@@ -0,0 +1,210 @@
...
+    exit 1
+  fi
+  eval "$var_name='$version'"
+}
+commit_changes=false
+create_pr=false
+# Parse named arguments
+while [[ "$#" -gt 0 ]]; do
+  case $1 in
+    --current-beta-core) current_beta_core="$2"; shift ;;
```
andrzej-stencel, line 43: Thinking about this more, do we actually even need the "current" versions from the user? Seems that we can read the current versions from the manifest files, e.g. with `awk`:

```console
$ current_beta_core=$(awk '/^.*go\.opentelemetry\.io\/collector\/.* v0/ {print $4; exit}' distributions/otelcol/manifest.yaml)
$ echo $current_beta_core
v0.111.0
```

andrzej-stencel (commented):
The script in the current form does not update the following places:

- `otelcol_version` and `version` in `manifest.yaml` files
- `OTELCOL_BUILDER_VERSION` in `Makefile`

I think the script should use the `--next-beta-core` version for the `OTELCOL_BUILDER_VERSION` in `Makefile` and for the `otelcol_version` field in manifests. As for the `version` field in manifest, it's probably max(next_beta_core, next_beta_contrib). Or we could add a new parameter to the script and action to specify it.

jackgopack4:
successful release PR in my fork: https://github.com/jackgopack4/opentelemetry-collector-releases/pull/21

jackgopack4:
@andrzej-stencel @mx-psi waiting for checks to run but I feel better about this attempt. incorporated feedback from code review and added changelog step

.github/workflows/scripts/bump-versions.sh
jackgopack4: got rid of the backup logic so this problem should be solved

jackgopack4:
> The script in the current form does not update the following places:
> 
> * `otelcol_version` and `version` in `manifest.yaml` files
> * `OTELCOL_BUILDER_VERSION` in `Makefile`
> 
> I think the script should use the `--next-beta-core` version for the `OTELCOL_BUILDER_VERSION` in `Makefile` and for the `otelcol_version` field in manifests. As for the `version` field in manifest, it's probably max(next_beta_core, next_beta_contrib). Or we could add a new parameter to the script and action to specify it.

addressed both of these

CHANGELOG.md
jackgopack4, line 1: required to make this work. Can remove from this PR once https://github.com/open-telemetry/opentelemetry-collector-releases/pull/775 is merged

andrzej-stencel (commented):
Thank you for your patience with my comments @jackgopack4. Just a couple small things and this will be ready to merge.

CHANGELOG.md
jackgopack4: yes, done

.github/workflows/update-version.yaml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Run bump-versions.sh
+        run: |
+          chmod +x .github/workflows/scripts/bump-versions.sh && \
```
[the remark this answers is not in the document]
jackgopack4: yes I didn't realize git makes it executable by default; fixed

andrzej-stencel (approved):
Thanks Jack!

PR #685, opened 2024-10-02
downgrade to go1.22 until there's a fix for timer

codeboten:
Proposing to downgrade the version of go used to produce the binary to give implementing a solution for https://github.com/open-telemetry/opentelemetry-collector/issues/11332 more time without impacting end users of the binary.


mx-psi (approved):

andrzej-stencel (approved):

codeboten:
This appears is no longer necessary as the next release will be built with 1.23.2 which includes a fix for the timer issue

Issue #668, opened 2024-09-19
Move back to `latest` for goreleaser


mx-psi:
Because of goreleaser/goreleaser/issues/5138, CI is currently broken when using goreleaser v2.3.0+.

#667 fixes this temporarily for the next release

jade-guiton-dd:
Are we sure we want to move back to latest instead of upgrading the version periodically? Manually upgrading would prevent the CI unexpectedly breaking on main like in #667 and #676.

mx-psi:
Could we automate the bumping? For other stuff like Github Actions we use dependabot/renovate bot, but here I don't think there is an equivalent tool that is ready for this. Maybe we can ask upstream on the goreleaser repositories for advice?

mx-psi:
@mowies Do you have any links/examples explaining how this would be done with renovate?

mowies:
Something like this should work with renovate. It's a [custom regex manager](https://docs.renovatebot.com/modules/manager/regex/) that matches the goreleaser action's input for the version. It could probably be done a little nicer and more resilient but it should work I think.
```
    {
      "fileMatch": [
        "(^|\\/).github\\/.*\\.ya?ml$"
      ],
      "matchStrings": [
        "goreleaser\\/goreleaser-action@[\\S\\s]+version: (?<currentValue>.*?)$"
      ],
      "depName": "github.com/goreleaser/goreleaser",
      "datasourceTemplate": "github-releases"
    }
```

mowies:
@mx-psi if you could take of setting up renovate in this repo, then i'd be happy to contribute the needed config changes :)

mx-psi:
I believe you will have to file an issue similar to this one: open-telemetry/community/issues/1792 but for this repository. I am happy to voice my support in such an issue.

mowies:
Cool! I filed the issue: https://github.com/open-telemetry/community/issues/2403

mx-psi:
Done!

mowies:
PR is open for this: https://github.com/open-telemetry/opentelemetry-collector-releases/pull/728

mowies:
seems like the regex I added still has some issues. I'll do a follow up PR

mowies:
@mx-psi maybe this PR fixes it: https://github.com/open-telemetry/opentelemetry-collector-releases/pull/733

mx-psi:
Was this fixed by #733?

mowies:
[...]

PR #689, opened 2024-10-04
[chore] Update goreleaser used in CI to v2.3.2

jade-guiton-dd:
The Github actions previously broke twice because of the release of goreleaser v2.3. I worked around this in PRs #667 and #676 by pinning the version used to v2.2.

After investigating, the causes of breakage were the following:

- In v2.3, the amd64 architecture got a version specifier, which is "v1" by default, so the output directory changed from just "linux_amd64" to "linux_amd64_v1". Our PR workflow takes the output of the linux_amd64 build tests and uses it in the "package tests" job. Because of the change in directory name, the build output could not be found, the artifact upload failed silently, and the package tests failed loudly as they could not find packages to test.

- In v2.3, the arm architecture changed its default version from "7" to "6". Our release CI never specifies this parameter when calling goreleaser, so it was now instructed to build for the arm_6 architecture, which we have no builds for in `.goreleaser.yaml` (only for arm_7). The prepare step thus failed silently, and the release step failed loudly as it can't find the artifacts for arm_7.

This PR fixes these bugs and updates the goreleaser version to v2.3.2, by updating the expected directory name for the linux_amd64_v1 build outputs, and by specifying the $GOARM version parameter.

If we expect no such breakage to happen again, this would allow us to go back to using the latest version of goreleaser at all times, as suggested in #668.

I haven't tested the release CI yet for lack of a goreleaser pro key.


songy23 (commented):
Does this need to wait for https://github.com/open-telemetry/opentelemetry-collector-releases/pull/672?

jade-guiton-dd:
@songy23 I don't think so. That PR updates the version of goreleaser used in the tool that generates the `.goreleaser.yaml` files, but upgrading to v2.3.2 doesn't seem to require changes on that front, only in the CI workflow files.

songy23 (approved):

mx-psi (approved):

PR #702, opened 2024-10-22
fix: add missing http ports to dockerfile expose statements

### LINKED ISSUE #692: Container image doesn't export http receiver 4318
The Perl OpenTelemetry::Exporter::OTLP module currently only supports the http protocol and not grpc.
The [config](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/distributions/otelcol-contrib/config.yaml#L17) enables the http receiver but the [Dockerfile](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/distributions/otelcol-contrib/Dockerfile#L14) doesn't expose it.
Why?

mowies:
This PR adds missing HTTP ports to the Dockerfile `EXPOSE` statements for otelcol and otelcol-contrib.

Fixes #692


mx-psi (approved):
Thank you!

PR #708, opened 2024-10-25
Update GoReleaser templating to support all collector distros

jackgopack4:
Per https://github.com/open-telemetry/opentelemetry-collector-releases/issues/716 we do not currently generate the Goreleaser configurations for `otelcol-otlp` and `otelcol-k8s` due to different requirements in default files, configurations, builds, etc.
This PR adds conditional logic to successfully generate these goreleaser templates.

Anyone can verify this change by downloading "main" branch, applying my changes to `Makefile` and `cmd/goreleaser/internal/configure.go`, running `$ make generate-goreleaser` and see the minute formatting changes that get applied to the goreleaser templates.

The pipeline tests should ensure that these binaries still build and release successfully but of course it should be verified after merge with v0.113.0 or v0.114.0

Updates to the goreleaser template will also likely be required to close https://github.com/open-telemetry/opentelemetry-collector-releases/issues/618 so I figured it was a good idea to get all four distros templated first


cmd/builder/.goreleaser.yml
```diff
@@ -8,6 +8,7 @@ version: 2
 builds:
   - flags:
       - -trimpath
+      - -buildmode=pie
```
jackgopack4, line 11: this change is safe to make because we only currently build/release ocb for platforms that support Internal Linking (buildmode=pie is successful even when CGO is disabled)

Makefile
```diff
@@ -40,7 +39,7 @@ ifeq (, $(shell command -v ocb 2>/dev/null))
 	[ "$${machine}" != x86_64 ] || machine=amd64 ;\
 	echo "Installing ocb ($${os}/$${machine}) at $(OTELCOL_BUILDER_DIR)";\
 	mkdir -p $(OTELCOL_BUILDER_DIR) ;\
-	CGO_ENABLED=0 go install -trimpath -ldflags="-s -w" go.opentelemetry.io/collector/cmd/builder@v$(OTELCOL_BUILDER_VERSION) ;\
+	CGO_ENABLED=0 go install -trimpath -ldflags="-s -w" -buildmode=pie go.opentelemetry.io/collector/cmd/builder@v$(OTELCOL_BUILDER_VERSION) ;\
```
jackgopack4, line 42: this change is safe as ocb is only built and used on platforms that support Internal Linking

jackgopack4:
@TylerHelmuth I think you've worked with the goreleaser templating recently; would you or someone else mind taking a look? Thanks.

jackgopack4:
@open-telemetry/helm-approvers @open-telemetry/operator-approvers appreciate some eyes on this change, thank you

swiatekm:
The templating changes look fine to me, but having them mixed with the PIE changes makes them harder to review. I'd prefer to merge #726 first. I also don't think we should proceed without @TylerHelmuth's review here.

jackgopack4:
> The templating changes look fine to me, but having them mixed with the PIE changes makes them harder to review. I'd prefer to merge #726 first.

Agreed, removed those changes.

cmd/goreleaser/internal/configure.go
```diff
@@ -67,26 +72,42 @@ func Builds(dist string) []config.Build {
 // Build configures a goreleaser build.
 // https://goreleaser.com/customization/build/
 func Build(dist string) config.Build {
+	var goos []string
+	var archs []string
+	var ignore []config.IgnoredBuild
+	var armVersions []string
+	flags := []string{"-trimpath"}
+	if dist == "otelcol-k8s" {
```
TylerHelmuth, line 80: Can you add this name as a const.

cmd/goreleaser/internal/configure.go
```diff
@@ -67,26 +72,42 @@ func Builds(dist string) []config.Build {
...
+		armVersions = ArmVersions
+	}
 	return config.Build{
 		ID:     dist,
 		Dir:    "_build",
 		Binary: dist,
 		BuildDetails: config.BuildDetails{
 			Env:     []string{"CGO_ENABLED=0"},
-			Flags:   []string{"-trimpath"},
+			Flags:   flags,
```
TylerHelmuth, line 104: It doesn't look like the `flags` variable is used elsewhere yet so this can stay as `[]string{"-trimpath"},`

cmd/goreleaser/internal/configure.go
```diff
@@ -30,9 +30,14 @@ import (
...
 
 var (
-	ImagePrefixes = []string{"otel", "ghcr.io/open-telemetry/opentelemetry-collector-releases"}
-	Architectures = []string{"386", "amd64", "arm", "arm64", "ppc64le", "s390x"}
-	ArmVersions   = []string{"7"}
+	ImagePrefixes      = []string{"otel", "ghcr.io/open-telemetry/opentelemetry-collector-releases"}
+	Architectures      = []string{"386", "amd64", "arm", "arm64", "ppc64le", "s390x"}
+	ArmVersions        = []string{"7"}
+	DefaultConfigDists = map[string]bool{"otelcol": true, "otelcol-contrib": true}
+	MSIWindowsDists    = map[string]bool{"otelcol": true, "otelcol-contrib": true, "otelcol-otlp": true}
```
TylerHelmuth, line 37: using const for this dist names wherever we can would be good in case we change the names in the future.

cmd/goreleaser/internal/configure.go
jackgopack4: set flags as a const set to `-trimpath` or set a const equal to `otelcol-k8s`?

TylerHelmuth: const equal to `otelcol-k8s`.  Lets use constants for all distro names in this generation process.

cmd/goreleaser/internal/configure.go
jackgopack4: are you saying define `otelcol`, `otelcol-contrib`, etc. as consts and then refer to those in the slices or maps here?

TylerHelmuth: Yes

TylerHelmuth (approved):

PR #727, opened 2024-11-06
remove go flag from distribution build script

jackgopack4:
in https://github.com/open-telemetry/opentelemetry-collector/pull/11576 we removed flag `--go` but this script will fail if builder v0.113.0 is used to build and release v0.113.0 distributions.

Pipeline build and release should be sufficient to test this script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ ci: check build
 check: ensure-goreleaser-up-to-date
 
 build: go ocb
-	@./scripts/build.sh -d "${DISTRIBUTIONS}" -b ${OTELCOL_BUILDER} -g ${GO}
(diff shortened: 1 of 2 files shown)


jackgopack4:
@djaglowski you'll likely want to merge this before preparing the v0.113.0 release

TylerHelmuth (commented):
Do we need to update out ocb config files?

jackgopack4:
doesn't look like it; we don't set `otelcol_version` in the distribution `manifest.yaml` files and they build fine locally with ocb v0.113.0, both as-is and when updating core component versions to v0.113.0/v1.19.0 (even leaving contrib at v0.112.0 since that release isn't published yet)

PR #728, opened 2024-11-06
Set up default renovate config with regex to handle goreleaser-pro updates

mowies:
This PR takes the renovate config from https://github.com/open-telemetry/opentelemetry-collector (as I found this to be mostly also applicable here, maybe a bit overkill still) and adds a custom regex manager that handles updates of the `version` field for GoReleaser inside the goreleaser-action usages.
Example of an occurrence that this should cover: https://github.com/open-telemetry/opentelemetry-collector-releases/blob/11530fd416496525d0a026be319feb442e0add3b/.github/workflows/base-release.yaml#L83-L88

Hopefully, this works for #668 but renovate config updates are hard to test before merging. The local setup didn't scream, which is a good sign. I will close the issue manually once I'm sure that this PR fixes it and otherwise provide follow-up PRs.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,7 +1,102 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": [
-    "config:recommended"
+  "labels": [
+    "renovatebot",
+    "dependencies"
+  ],
+  "constraints": {
+    "go": "1.22.0"
+  },
+  "schedule": ["every tuesday"],
+  "extends": ["config:recommended"],
+  "packageRules": [
+    {
+      "matchManagers": ["gomod"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,6 +1,89 @@
...
+      "customType": "regex",
+      "fileMatch": [
+        "(^|\\/).github\\/.*\\.ya?ml$"
+      ],
+      "matchStrings": [
+        "goreleaser\\/goreleaser-action@[\\S\\s]+version: (?<currentValue>.*?)$"
+      ],
+      "depNameTemplate": "github.com/goreleaser/goreleaser-pro",
+      "datasourceTemplate": "github-releases"
+    }
```
mowies, line 85: This is the part that's new to update the goreleaser-pro version in the github action.
The rest of the PR is the copied renovate config from collector-core repo.

mx-psi (approved):
Looks reasonable to me, let's merge it and try it out!

Issue #662, opened 2024-09-16
container user 10001 is group root. 


0x6f677548:
(originally reported at https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35179)

All distributions are running the container as user 10001, but no group was assigned, **meaning that this user will be assigned root group.**

Actual config:
```
ARG USER_UID=10001
USER ${USER_UID}:${USER_GID}
```

My suggestion of fix:
```
ARG USER_UID=10001
ARG USER_GID=10001
USER ${USER_UID}:${USER_GID}
```

As pointed by @rogercoll, based on [dockerfile documentation](https://docs.docker.com/reference/dockerfile/#user):

> When the user doesn't have a primary group then the image (or the next instructions) will be run with the root group.

[security best-practices](https://docs.docker.com/build/building/best-practices/):

> Consider an explicit UID/GID.
> Users and groups in an image are assigned a non-deterministic UID/GID in that the "next" UID/GID is assigned regardless of image rebuilds. So, if it’s critical, you should assign an explicit UID/GID.
> 

This can be reproduced by shelling "id" on a modified image:
```

~ $ id
uid=10001 gid=0 groups=0
```

Modified docker image to include a shell and "id":
```
FROM alpine:latest AS builder

# Install a statically linked shell and the necessary binaries
RUN apk add --no-cache busybox-static
RUN apk add --no-cache coreutils acl attr

FROM otel/opentelemetry-collector-contrib:latest AS prep

FROM scratch

# Copy the shell executable from the builder stage
COPY --from=builder /bin/busybox.static /bin/sh
# Copy id binary from the builder stage
COPY --from=builder /usr/bin/id /bin/id

# Copy required shared libraries
COPY --from=builder /lib/ld-musl-x86_64.so.1 /lib/ld-musl-x86_64.so.1
COPY --from=builder /lib/libc.musl-x86_64.so.1 /lib/libc.musl-x86_64.so.1
COPY --from=builder /lib/libcrypto.so.3 /lib/
COPY --from=builder /lib/libacl.so.1 /lib/
COPY --from=builder /lib/libattr.so.1 /lib/
COPY --from=builder /lib/libutmps.so.0.1 /lib/
COPY --from=builder /lib/libskarnet.so.2.14 /lib/
COPY --from=builder /lib/libutmps.so.0.1 /lib/

[...]

PR #738, opened 2024-11-18
chore(Dockerfile): set USER_GID alongside USER_UID

rogercoll:
Fixes https://github.com/open-telemetry/opentelemetry-collector-releases/issues/662


iblancasa (commented):
I think we should create a changelog for this because I'm not sure if this change can cause unexpected issues in some environments.

rogercoll:
> I think we should create a changelog for this because I'm not sure if this change can cause unexpected issues in some environments.

Is there any related documentation guidelines? I could not find any changelog file in the repo

swiatekm:
I think we use the autogenerated Github release notes in this repo. So no separate changelog necessary.

swiatekm (approved):

iblancasa (approved):

mowies:
would it not be better to add an actual collector changelog entry as a backport in the other repo(s)? This seems important enough to warrant that

TylerHelmuth (approved):

rogercoll:
> would it not be better to add an actual collector changelog entry as a backport in the other repo(s)? This seems important enough to warrant that

There is a changelog entry in the contrib v0.114.0 release https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.114.0. The core collector does not have a container Dockerfile, maybe we could add a PR with just a changelog entry referencing this PR?

ChrsMark:
Could we mention what kind of testing has been performed here? I'm mostly worried about any potential issues this can cause to existing users. I would expect that users using the upstream Helm charts and/or Operator to not be affected. @rogercoll is this verified?

Based on the above ,+1 on surfacing this change as much as possible to the release notes.

mowies:
> There is a changelog entry in the contrib v0.114.0 release [open-telemetry/opentelemetry-collector-contrib@`v0.114.0` (release)](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.114.0).

perfect

> The core collector does not have a container Dockerfile, maybe we could add a PR with just a changelog entry referencing this PR?

the core collector also has one but it's only in the releases repo. In any case, the dockerfile in the contrib repo is only used for testing and not for actual releases. The release dockerfiles are all in the releases repo. So I think it would also be good to have a changelog entry for the core collector just in a standalone PR.

rogercoll:
> Could we mention what kind of testing has been performed here? I'm mostly worried about any potential issues this can cause to existing users. I would expect that users using the upstream Helm charts and/or Operator to not be affected. @rogercoll is this verified?
> 
> Based on the above ,+1 on surfacing this change as much as possible to the release notes.

I don't think there is any mechanism to test the upstream Helm Charts with an unreleased version of the Collector.  But I published an image from this branch and manually tested the OpenTelemetry Demo Chart `helm install my-otel-demo open-telemetry/opentelemetry-demo --set opentelemetry-collector.image.repository="rcolles/otelcol-contrib" --set opentelemetry-collector.image.tag="nonrootgroup"`

rogercoll:
> So I think it would also be good to have a changelog entry for the core collector just in a standalone PR.

Sounds good to me, not sure if these situation has happened before, but here is the changelog PR https://github.com/open-telemetry/opentelemetry-collector/pull/11829

atoulme:
This repository now supports chloggen, please add a changelog entry to the PR

rogercoll:
> This repository now supports chloggen, please add a changelog entry to the PR

Great!  added https://github.com/open-telemetry/opentelemetry-collector-releases/pull/738/commits/d4eb034a3282d74cea9072f22b889a384a73d1f1

jackgopack4 (approved):
LGTM, thanks

mx-psi (approved):

PR #748, opened 2024-11-24
Update otelcol-contrib.service

KieranP:
Add ExecReload command to reload configs without restarting the opentelemetry process (to prevent data loss)

Taken from https://last9.io/blog/hot-reload-for-opentelemetry-collector/


distributions/otelcol-contrib/otelcol-contrib.service
```diff
@@ -5,6 +5,7 @@ After=network.target
 [Service]
 EnvironmentFile=/etc/otelcol-contrib/otelcol-contrib.conf
 ExecStart=/usr/bin/otelcol-contrib $OTELCOL_OPTIONS
+ExecReload=/bin/kill -HUP $MAINPID
```
mowies, line 8: please also add this to otelcol and otelcol-otlp

KieranP: @mowies Done :-)

Issue #756, opened 2024-11-28
Create follow up issue when nightly tests fail


mx-psi:
How will we be able to tell if there is an error? I think we should create a `release:blocker` issue if there is a failed test. We can do this in a follow up PR if you want to.

_Originally posted by @mx-psi in https://github.com/open-telemetry/opentelemetry-collector-releases/pull/666#discussion_r1860502491_

mowies:
The issue should ideally be created in -contrib, if that's not possible then in -releases

mowies:
pls assign this to me :)

PR #762, opened 2024-12-05
test: run linux package tests nightly against latest contrib main [new]

mowies:
Supersedes https://github.com/open-telemetry/opentelemetry-collector-releases/pull/760

Changes compared to the previously reverted PR:
- adapted the `base-release.yaml` to the changed gorelease setup for otelcol-contrib

Copied description from old PR:

This PR enables nightly testing of the otelcol-contrib linux packages based on the latest main branch version of otelcol-contrib built in https://github.com/open-telemetry/opentelemetry-collector-contrib.
This change was made possible by splitting the otelcol-contrib gorelease config into 2 separate files, one for building and one for packaging/archives/docker images/etc.
(The split goreleaser build is a goreleaser pro feature)

Using the split building approach was the only viable option to achieve this. I did lots of research on this, it was talked about in the Collector SIG meeting and I also [asked the GoReleaser folks for their help](https://github.com/goreleaser/goreleaser/issues/5234).

Consequences of this change:
- better testing
- slightly less maintainability of the goreleaser files and generator code for otelcol-contrib since they have special handling now

Fixes #659

PR #776, opened 2024-12-16
[chore] Create issue when nightly tests fail

mowies:
This PR adds issue creation when the nightly package tests fail. The issue will be created in the -contrib repo with the "release:blocker" label attached.

Fixes https://github.com/open-telemetry/opentelemetry-collector-releases/issues/756

### DIFF
--- a/.github/workflows/base-package-tests.yaml
+++ b/.github/workflows/base-package-tests.yaml
@@ -9,6 +9,11 @@ on:
       distribution:
         required: true
         type: string
+      issue-on-failure:
+        required: false
+        type: boolean
+        default: false
+        description: "Set to true if a GH issue should be generated upon failure"
 
 jobs:
   package-tests:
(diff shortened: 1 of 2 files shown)


mowies:
blocked by #762

.github/workflows/base-package-tests.yaml
```diff
@@ -28,3 +33,29 @@ jobs:
 
       - name: Test ${{ matrix.type }} package
         run: ./scripts/package-tests/package-tests.sh ./otelcol*-SNAPSHOT-*_linux_amd64.${{ matrix.type }} ${{ inputs.distribution }}
+
+  create-issue:
+    name: Create GitHub Issue
+    runs-on: ubuntu-24.04
```
songy23, line 39:
```suggestion
    runs-on: ubuntu-22.04
```

.github/workflows/base-package-tests.yaml
```diff
@@ -28,3 +33,29 @@ jobs:
...
+          echo "* Triggered by: Nightly scheduled build" >> nightly-test-failure.md
+          echo "" >> nightly-test-failure.md
+          echo "Note: This issue was auto-generated from [base-package-tests.yaml](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/.github/workflows/base-package-tests.yaml)" >> nightly-test-failure.md
+
+      - name: Create issue on failure
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PAT }}
+        run: |
+          gh issue create \
+            --repo open-telemetry/opentelemetry-collector-contrib \
```
songy23, line 58: question: why always in contrib?

mowies: that was requested by @mx-psi in the original comment

mx-psi: I guess not all changes would be things to fix in contrib, but my bet is that most of them would be, thus why I suggested putting them in contrib. Since this is easy to change and does not have an impact on end users, we can start with this and see how it goes

mx-psi (approved):

PR #795, opened 2025-01-13
Add Windows build and refactor goreleaser config generation

### LINKED ISSUE #339: Release collector Docker images as packages for Windows
The releases repo presently includes Docker images as [packages](https://github.com/open-telemetry/opentelemetry-collector-releases/pkgs/container/opentelemetry-collector-releases%2Fopentelemetry-collector-contrib), but these are available only for Linux operating systems.

I'm starting to have more use cases where I'd like to run collector on Windows in a container (on a K8s node). I believe having these images available as part of the official releases would be beneficial to more user.

I'm not sure about all the caveats and gotchas that might exists with releasing images for Windows (I think at least we might have to have separate images for Windows 2019 and Windows 2022), but I believe it should be doable. If the maintainers are in favour of adding them, I'd be happy to help with the necessary work.

douglascamata:
This should fix #339, while at the same time moving the code for generating the `goreleaser` configuration for the distros into a more declarative and easy-to-extend form. 

The Windows build will be producing images based on Windows 2019 and 2022. 

## Why so many changes? 😱

I am completely aware that it represents a big rewrite of this part of code. I tried a few different approaches with the intent of having a minimum change that would allow me to support Windows, but the amount of changes needed always ended up snowballing quickly. Some examples of the reasons for this "snowballing" of code are (non-exhaustive list):

- Handling arm vs arm/v7. 
- Handling Windows' requirement of different base images depending on the host OS Windows version.
- "Build-only" distro for core distro.

This might not the best code for this job too, but at least I believe it sets up a good structure that we can build upon and iterate on to improve.

I'm very open to hear your feedback and see how we can move this forward!

I can join the Collector SIG call on 2025-01-15 to talk about this if any maintainer finds it more productive.

## Questions ⍰

If anyone, please, could help me out in testing that release process works fine with the update goreleaser configuration, it would be aaaaamazing! I don't feel super confident without testing this, of course.


andrzej-stencel:
> Add Windows build and refactor goreleaser config generation

Thanks for the contribution Matej!

Can you implement the refactoring in a separate PR from the functional change? This should make code review much more effective.

douglascamata:
> > Add Windows build and refactor goreleaser config generation
> 
> Thanks for the contribution Douglas!
> 
> Can you implement the refactoring in a separate PR from the functional change? This should make code review much more effective.

Sure thing. I'll open a separate PR for the refactor, cross-reference this one and close it so that we can focus on the new one.

douglascamata:
Opened #797, which should be only the refactor but without the Windows containers parts, @andrzej-stencel. Closing this.

PR #800, opened 2025-01-17
fix: fix Linux package scripts

### LINKED ISSUE #799: Systemd service stopped and disabled after rpm package update
I recently updated our otelcol-contrib rpm package from version 0.112.0 to 0.117.0 and noticed that after the update the otelcol-contrib systemd service was stopped and disabled.

As you can see in https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#ordering, is first executed the **postinstall** (post) of the previous package version (4.) and then the **preuninstall** (preun) of the new package version (9.).

In the here generated rpm package is first the otelcol-contrib service enabled and started in the **postinstall** of the previous version, and then stopped and disabled in the **preuninstall** of the new version.

The deb package seams to be not affected, because regarding https://wiki.debian.org/MaintainerScripts#Upgrading is there first executed the **preuninstall** (prerm) of the previous version and then the **postinstall** (postinst) of the new version.

I will later create a PR for this.

netsandbox:
fixes #799

### DIFF
--- a/distributions/otelcol-otlp/preremove.sh
+++ b/distributions/otelcol-otlp/preremove.sh
@@ -14,7 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-if command -v systemctl >/dev/null 2>&1; then
-    systemctl stop otelcol-otlp.service
-    systemctl disable otelcol-otlp.service
+if [ "$1" != "1" ]; then
+    if command -v systemctl >/dev/null 2>&1; then
+        systemctl stop otelcol-otlp.service
+        systemctl disable otelcol-otlp.service
+    fi
 fi
--- a/.chloggen/fix_linux_package_scripts.yaml
+++ b/.chloggen/fix_linux_package_scripts.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: packaging
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix Linux package scripts
+
+# One or more tracking issues or pull requests related to the change
+issues: [799]
+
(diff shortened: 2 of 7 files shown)


distributions/otelcol-otlp/preremove.sh
```diff
@@ -14,7 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-if command -v systemctl >/dev/null 2>&1; then
-    systemctl stop otelcol-otlp.service
-    systemctl disable otelcol-otlp.service
+if [ "$1" != "1" ]; then
```
atoulme, line 17: What does this check on?

atoulme:
Please include a changelog. Any way to test those changes?

netsandbox:
I have now added a changelog.

If the build rpm package works can be tested on a RHEL based OS host.

Actually the main fix from this MR, that the service is stopped and disabled after a rpm package update, can't be tested now, because when you install a package release with the fix from this MR, the package **preuninstall** script from the **previous** release will still stop and disable the service (see details on this in the linked issue #799).
You will only see if this works when you install a release, after the release with this MR fix.

atoulme:
Makes sense, that's an unhappy confluence of events, we should ideally add tests to test lifecycle after this release.

netsandbox:
I was also a little surprised that no one else ever noticed this since now.

It would make sense to create a release after this MR is merged, to get a fixed version out as soon as possible

atoulme (approved):

MovieStoreGuy (approved):

mowies (commented):
will your changes also work with the .deb package for debian based systems?
the shell scripts are also used for that package

netsandbox:
> will your changes also work with the .deb package for debian based systems? the shell scripts are also used for that package

Yes, the changes work also for the Debian packages.
As the Debian packages use a different script order, they are not affected by this problem.

We just don't run the service stop and disable when `$1` is `1`, and this is only the case if the rpm package is upgraded. Debian package install never provides `1` for `$1` to a prerm script, see https://wiki.debian.org/MaintainerScripts

mowies (approved):

codeboten (approved):

Issue #824, opened 2025-02-12
Use Go 1.24 for official builds


songy23:
Depends on https://github.com/open-telemetry/opentelemetry-collector/issues/12362 and https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/37865.

Once core and contrib add tests on Go 1.24 we can migrate to Go 1.24 for official builds in this repo.

PR #827, opened 2025-02-12
bump version to go1.24

codeboten:
With the release of go 1.24, the versions used must be bumped. Note that the docker image for 1.24 is not yet published.

Fixes https://github.com/open-telemetry/opentelemetry-collector-releases/issues/824


songy23 (commented):
need a changelog?

douglascamata:
We gotta wait for the 1.24 image to be published otherwise this will break all following PR builds.

codeboten:
@douglascamata the image is present, and the changelog is there as well

mowies (approved):
LGTM, go.mod versions were already at 1.23 (oldest supported version) that's why you didn't bump them right?

PR #837, opened 2025-02-25
Add Docker test pipeline

mowies:


mx-psi (approved):
Looks reasonable, it may flake if the Collector takes too long to start up but I guess it's also good to be aware of that possibility. Thanks!!

PR #850, opened 2025-03-04
[chore] Remove unneeded build combinations and inconsistencies

### LINKED ISSUE #838: Remove unneeded OS/Arch combinations
I found a few build combinations that are run in our release and CI pipelines that never actually produce any artifacts and are just needlessly taking up resources. This makes builds unreliable sometimes because the existence of common artifacts like container images can't be taken for granted in the pipelines.

Combinations I found so far (but maybe there are more):
- darwin/ppc64le
- windows/ppc64le

mowies:
### This PR
- removes the build combinations darwin/ppc64le and windows/ppc64le because they are not supported by Go and no artifacts of any sort are generated from them.
- makes pipeline naming more consistent
  - Continuous Integration -> CI
  - renamed builder pipelines to adhere to same naming scheme as collector distros

Fixes #838

### DIFF
--- a/.github/workflows/base-ci-goreleaser.yaml
+++ b/.github/workflows/base-ci-goreleaser.yaml
@@ -41,6 +41,10 @@ jobs:
             GOARCH: arm
           - GOOS: windows
             GOARCH: s390x
+          - GOOS: darwin
+            GOARCH: ppc64le
+          - GOOS: windows
+            GOARCH: ppc64le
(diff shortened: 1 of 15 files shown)


evan-bradley (approved):
Looks good to me. I validated the allowed GOOS and GOARCH combinations with this list: https://go.dev/doc/install/source#environment

PR #860, opened 2025-03-06
[chore] Pull systemd service logs on package test failure

douglascamata:
This should help us identify issues introduced in the release build process related to RPM/deb packaging.

### DIFF
--- a/scripts/package-tests/package-tests.sh
+++ b/scripts/package-tests/package-tests.sh
@@ -50,6 +50,12 @@ podman run --name "$container_name" -d "$image_name"
 $container_exec systemctl is-system-running --quiet --wait
 install_pkg "$container_name" "$PKG_PATH"
 
+# If we got to this point, we might need to check the logs of the systemd service
+# when it's not properly active. This is added as a trap because the check
+# for service status below will return an error exitcode if the service is 
+# not active, triggering the end of this script because of the shell option `-e`
+trap '$container_exec journalctl -u "$SERVICE_NAME" || true' EXIT
+
 # ensure service has started and still running after 5 seconds
 sleep 5
 echo "Checking $SERVICE_NAME service status ..."
--- a/scripts/package-tests/package-tests.sh
+++ b/scripts/package-tests/package-tests.sh
@@ -14,6 +14,25 @@ DISTRO="${2:-}"
 SERVICE_NAME=$DISTRO
 PROCESS_NAME=$DISTRO
 
+# Global array for trap functions
+TRAP_FUNCS=()
+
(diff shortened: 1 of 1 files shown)


mowies (commented):
does having 2 traps work? 🤔

douglascamata:
@mowies good catch! I just learned that we can only have one trap per signal. I think I found a workaround that is a nice: creating a trap to call functions from a list, then append more functions to the list as necessary. Let me see runs some tests and see if it'll work.

douglascamata:
@mowies I pushed a fix with an approach to allow us to stack multiple functions to be "trapped". Let me know what you think.

Tested locally with the script:

```bash
#!/bin/bash

# Copyright The OpenTelemetry Authors
# SPDX-License-Identifier: Apache-2.0

set -euo pipefail

TRAP_FUNCS=()

add_trap_func() {
	TRAP_FUNCS+=("$1")
}

run_traps() {
	if [ "${#TRAP_FUNCS[@]}" -gt 0 ]; then
		for ((i = ${#TRAP_FUNCS[@]} - 1; i >= 0; i--)); do
			"${TRAP_FUNCS[i]}"
		done
	fi
}
trap 'run_traps' EXIT

my_echo() {
	echo "my echo"
}

my_other_echo() {
	echo "my other echo"
}

add_trap_func my_echo
echo "stacked first"

add_trap_func my_other_echo
echo "stacked second"
```

And got:

```
╰─ ./test.sh
stacked first
stacked second
my other echo
my echo
```

douglascamata:
A nice "plus" of this approach is that it only works if you define the traps as functions, which allows linting tools like shellcheck and others to run and boost confidence on the code. Previously the trapped code was inside a string that was treated just like any random string and could hide bugs until runtime.

mowies:
sounds good to me :)

atoulme:
Can we make the whitespace changes separate? Can you approve the PR if you're ok with it @mowies ?

douglascamata:
@atoulme you mean in a separate PR? The formatting is already separated in its own commit.

mowies (approved):

douglascamata:
Hey @atoulme, can you check my previous reply regarding the commit with the whitespaces change? Just want to clarify whether "making the whitespace changes separate" means a separate PR or commit, because it's already in a separate commit.

jackgopack4:
I think he means separate PR; just makes it easier to review.

douglascamata:
@jackgopack4 done.

jackgopack4 (approved):
Looks good, thanks for splitting out the linter changes

mowies (approved):

PR #872, opened 2025-03-12
[feat] Add Windows 2019 and 2022 container image (amd64 only)

douglascamata:
This PR adds support for releasing Windows 2019 and 2022 container images for the `amd64` architecture. 

As a reminder, [`windows/amd64` is already a supported platform (tier 2) in the Collector](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/platform-support.md#tier-2--secondary-support). 

Another worthy note: there are no Windows containers image produced by Microsoft on the i386 architecture. 

First of all, this was only possible after reading a lot of material about building Windows container images (https://lippertmarkus.com/2021/11/30/win-multiarch-img-lin/, https://tobiasfenster.io/building-docker-images-for-multiple-windows-server-versions-using-self-hosted-github-runners) and seeing some examples of other projects using GoReleaser that also build Windows container images (https://github.com/testcontainers/moby-ryuk/pull/52, for example). 

As byproduct of this pull request, I started [a discussion in GoReleaser to talk about my wishlist to improve support for building Windows container images with the "Split and Merge" strategy](https://github.com/orgs/goreleaser/discussions/5638). Some of the small hacks in this PR could be remove in the future as outcome of this conversation. I'll keep collaborating with the folks there to improve the support for this.

# Additional changes

- Fix slow cache restore of the `setup-go` Github Action in Windows: pretty self explanatory. The action was taking over 10 minutes to restore the cache. There are issues and PRs upstream to the action with a fix (see https://github.com/actions/setup-go/pull/515), but they weren't merged it.
- The GoReleaser `release` configuration now has `replace_existing_artifacts: true`. This is required because the split-and-merge build will try to upload some generic artifacts (like source code tarball) more than once. Without this setting on, when it tries to upload an artifact that already exists an error is returned and the build fails.
- The ID of the executable in the MSI installer template replaces dashes with underscore (`Id="{{ replace .Binary "-"  "_"}}.exe"`). This is required because now the MSI installer is built in a Windows node, which enforces that IDs contains only alphanumeric characters, dots, and underscores.

# Testing

I produced a test release in my fork using this code: https://github.com/douglascamata/opentelemetry-collector-releases/releases/tag/v0.121.0.

It contains 458 release artifacts. The upstream v0.121.0 contains 452. The new artifacts are the files for the Windows amd64 binary that now is built in the `otelcol-k8s` distro:

image

# Future work

## Support for `windows/arm64`

In the near feature we can add support for `windows/arm64`. I didn't add it right now for two main reasons:

- It's not a platform that is currently supported by the OpenTelemetry Collector. In fact, I tried to compile it and it didn't work because some dependencies do not compile on Windows arm64.
- Only Windows Server 2025 has arm64 images. Which means I would need to complicate this a bit more to skip arm64 for Windows 2019 and 2022.

## Windows multi-version + multi-arch images

It is possible to build a multi-version image for Windows. This means, for example, a single image manifest tagged as `otel/opentelemetry-collector-otlp:{{ .Version }}-windows-amd64` can include images for Windows 2019 and 2022.

This can be further extended by also including in the manifest images for all the architectures supported by each Windows version. We would end up with a manifest tagged like `otel/opentelemetry-collector-otlp:{{ .Version }}-windows` and potentially including support for Windows 2019, 2022, and 2025 -- including amd64 and arm64 for Windows 2025, if they will be ready then.

I didn't do this right now for two reasons:

- This PR would get bigger and it's already big enough.
- Windows containers are, so far, supported only on amd64 anyway.

This could be done in a follow up PR.

### Universal image (the dream)

It is possible to combine Windows multi-version + multi arch image with the Linux images. We would end up with a single manifest tagged as `otel/opentelemetry-collector-otlp:{{ .Version }}` that would work on all the supported Linux and Windows architectures and versions. 🤯

It's currently impossible to implement this though because of limitations in how GoReleaser handles image manifests. It currently has to load all the images going into a manifest and, as we know, Windows container image cannot be loaded on Linux and vice-versa.

### DIFF
--- a/.github/workflows/base-release.yaml
+++ b/.github/workflows/base-release.yaml
@@ -12,6 +12,10 @@ on:
       goarch:
         required: true
         type: string
+      runner_os:
+        required: false
+        type: string
+        default: ubuntu-24.04
 
 env:
   # renovate: datasource=github-tags depName=goreleaser-pro packageName=goreleaser/goreleaser-pro
@@ -21,26 +25,18 @@ jobs:
   prepare:
     strategy:
       matrix:
-        GOOS: ${{ fromJSON( inputs.goos) }}
-        GOARCH: ${{ fromJSON( inputs.goarch) }}
+        GOOS: ${{ fromJSON(inputs.goos) }}
+        GOARCH: ${{ fromJSON(inputs.goarch) }}
         exclude:
           - GOOS: darwin
             GOARCH: "386"
           - GOOS: darwin
             GOARCH: s390x
-          - GOOS: windows
-            GOARCH: arm64
-          - GOOS: darwin
-            GOARCH: arm
-          - GOOS: windows
-            GOARCH: arm
-          - GOOS: windows
-            GOARCH: s390x
           - GOOS: darwin
             GOARCH: ppc64le
-          - GOOS: windows
-            GOARCH: ppc64le
-    runs-on: ubuntu-24.04
+          - GOOS: darwin
+            GOARCH: arm
+    runs-on: ${{ inputs.runner_os }}
 
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
@@ -52,21 +48,27 @@ jobs:
       - uses: anchore/sbom-action/download-syft@f325610c9f50a54015d37c8d16cb3b0e2c8f4de0 # v0.18.0
 
       - uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
+        if: runner.os != 'Windows'
         with:
           platforms: arm64,ppc64le,linux/arm/v7,s390x
 
       - uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2 # v3.10.0
+        if: runner.os != 'Windows'
+
+        # Fix slow Go compile and cache restore
+        # See https://github.com/actions/setup-go/pull/515
+      - name: Fix slow setup-go cache restore in Windows
+        if: runner.os == 'Windows'
+        run: |
+          echo "GOCACHE=D:\gocache" | Out-File -FilePath $env:GITHUB_ENV -Append
+          echo "GOMODCACHE=D:\gomodcache" | Out-File -FilePath $env:GITHUB_ENV -Append
+          echo "GOTMPDIR=D:\gotmp" | Out-File -FilePath $env:GITHUB_ENV -Append
+          mkdir D:\gotmp
 
       - uses: actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34 # v5.3.0
         with:
           go-version: "~1.24"
           check-latest: true
-
-      - name: Setup wixl # Required to build MSI packages for Windows
-        if: ${{ matrix.GOOS == 'windows' && ( matrix.GOARCH == '386' || matrix.GOARCH == 'amd64') }}
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y wixl
 
       - name: Generate distribution sources
         run: make generate-sources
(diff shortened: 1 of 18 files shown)


mowies:
Fixes #339

mowies:
please add a changelog entry for this

douglascamata:
Changelog entry added, @mowies. I always leave it for the end and forget it. Sorry. 😅

distributions/otelcol-contrib/Windows.dockerfile
```diff
@@ -0,0 +1,12 @@
+# escape=`
+ARG WIN_VERSION=2019
+FROM mcr.microsoft.com/windows/nanoserver:ltsc${WIN_VERSION}
+
+COPY otelcol-contrib.exe ./otelcol-contrib.exe
+COPY config.yaml ./config.yaml
+
+ENV NO_WINDOWS_SERVICE=1
```
pjanotti, line 8: This environment variable has no effect anymore, it is safe to remove it.

distributions/otelcol-k8s/Windows.dockerfile
```diff
@@ -0,0 +1,15 @@
+# escape=`
+ARG WIN_VERSION=2019
+FROM mcr.microsoft.com/windows/nanoserver:ltsc${WIN_VERSION}
+
+COPY otelcol-k8s.exe ./otelcol-k8s.exe
+
+ENV NO_WINDOWS_SERVICE=1
```
pjanotti, line 7: Safe to remove

distributions/otelcol-otlp/Windows.dockerfile
```diff
@@ -0,0 +1,10 @@
+# escape=`
+ARG WIN_VERSION=2019
+FROM mcr.microsoft.com/windows/nanoserver:ltsc${WIN_VERSION}
+
+COPY otelcol-otlp.exe ./otelcol-otlp.exe
+
+ENV NO_WINDOWS_SERVICE=1
```
pjanotti, line 7: Safe to remove

distributions/otelcol/Windows.dockerfile
```diff
@@ -0,0 +1,12 @@
+# escape=`
+ARG WIN_VERSION=2019
+FROM mcr.microsoft.com/windows/nanoserver:ltsc${WIN_VERSION}
+
+COPY otelcol.exe ./otelcol.exe
+COPY config.yaml ./config.yaml
+
+ENV NO_WINDOWS_SERVICE=1
```
pjanotti, line 8: Safe to remove

.chloggen/add_windows_container_images.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: packaging
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Start producing images for Windows 2019 and 2022 on amd64
```
pjanotti, line 10:
```suggestion
note: Start producing Docker images for Windows 2019 and 2022 on amd64
```

.github/workflows/base-release.yaml
```diff
@@ -52,22 +48,28 @@ jobs:
...
+        if: runner.os != 'Windows'
         with:
           platforms: arm64,ppc64le,linux/arm/v7,s390x
 
       - uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2 # v3.10.0
+        if: runner.os != 'Windows'
+
+        # Fix slow Go compile and cache restore
+        # See https://github.com/actions/setup-go/pull/515
+      - name: Fix slow setup-go cache restore in Windows
```
pjanotti, line 60: Good to know about this: I was annoyed by it, but never dig into it, thanks!

.github/workflows/base-release.yaml
```diff
@@ -115,10 +117,14 @@ jobs:
         if: inputs.distribution == 'otelcol-contrib'
         run: mv distributions/otelcol-contrib/dist/**/* distributions/otelcol-contrib/artifacts/
 
-      - name: Show built or downloaded content
-        if: inputs.distribution == 'otelcol-contrib'
+      - name: Show built or downloaded content (Linux)
+        if: inputs.distribution == 'otelcol-contrib' && runner.os != 'Windows'
         run: ls -laR distributions/otelcol-contrib/artifacts
 
+      - name: Show built or downloaded content (Windows)
```
pjanotti, line 124: Have you tried `shell: bash` instead of using two different actions?

.chloggen/add_windows_container_images.yaml
mowies: :)
```suggestion
note: Start producing container images for Windows 2019 and 2022 on amd64
```

douglascamata: Going with @mowies suggestion to be more neutral.

.github/workflows/base-release.yaml
douglascamata: Good point. I didn't try it out with `bash`. I'll check whether it works, because indeed having repeated steps for different OSes doesn't feel good.

douglascamata: @pjanotti I could apply this suggestion in some parts of the workflow. 

In some steps, like `make generate-sources`, it didn't work. For instance, in this specific example it wasn't being able to find the `ocb` binary.

But it already saved us some duplication and conditionals.

pjanotti (approved):
LGTM

douglascamata:
Whatever happens, let's only merge this after 0.122.0 is safely released.

atoulme (approved):

PR #873, opened 2025-03-12
[chore] Add dockerhub metadata with links to readme files

### LINKED ISSUE #862: [cmd/builder] OCB DockerHub page does not have "overview"
The [OpenTelemetry Collector Builder DockerHub page](https://hub.docker.com/r/otel/opentelemetry-collector-builder) does not have an "overview".

This would make it easier for users to read more about how to install and configure these docker images.

I know this is supposed to be automatically linked to the README.md if the repository/docker push command is set up correctly; there should be a way to do this in goreleaser.

mowies:
Fixes #862

Especially the short descriptions need some improvement, but that can be done as a follow-up as well.

### DIFF
--- a/.chloggen/contrib-awscloudwatchmetricsencodingextension.yaml
+++ b/.chloggen/contrib-awscloudwatchmetricsencodingextension.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: awscloudwatchmetricstreamsencodingextension
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add the AWS CloudWatch Metric Streams encoding extension to the contrib distribution
+
+# One or more tracking issues or pull requests related to the change
+issues: [883]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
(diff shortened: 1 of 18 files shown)


distributions/otelcol/.goreleaser.yaml
```diff
@@ -324,3 +324,10 @@ monorepo:
   tag_prefix: v
 partial:
   by: target
+dockerhub:
+  - images:
+      - otel/opentelemetry-collector
+    description: OpenTelemetry Collector - otelcol
+    full_description:
+      from_url:
+        url: https://raw.githubusercontent.com/open-telemetry/opentelemetry-collector/refs/heads/main/README.md
```
jackgopack4, line 333: same question about using releases distro README vs. other repo, I'm open to either though

cmd/goreleaser/internal/configure.go
```diff
@@ -41,11 +41,17 @@ const (
...
-	winArchs          = []string{"386", "amd64", "arm64"}
-	winContainerArchs = []string{"amd64"}
-	darwinArchs       = []string{"amd64", "arm64"}
-	k8sArchs          = []string{"amd64", "arm64", "ppc64le", "s390x"}
+	baseArchs                 = []string{"386", "amd64", "arm", "arm64", "ppc64le", "s390x"}
+	winArchs                  = []string{"386", "amd64", "arm64"}
+	winContainerArchs         = []string{"amd64"}
+	darwinArchs               = []string{"amd64", "arm64"}
+	k8sArchs                  = []string{"amd64", "arm64", "ppc64le", "s390x"}
+	dockerHubFullDescriptions = map[string]string{
```
jackgopack4, line 49: very nice; in the future maybe there will be a way to tie the README URLs to a specific tag (like having a template variable that gets replaced by the github action) but I doubt these are changing that often so it's probably fine.

jackgopack4 (approved):
just curious about specific links, whether we should keep them consistent or not in terms of repo. Nice work, though. Thank you for picking this up.

atoulme (approved):

Issue #767, opened 2024-12-10
`postinstall.sh` should detect chrooted environments to prevent dpkg failures


bowling233:
### **Description**  
When installing `otelcol` in chrooted environments (e.g., `chroot` or `systemd-nspawn`), the installation fails because the `postinstall.sh` script attempts to run `systemctl` commands to start and enable systemd services.  

In these environments, `systemd` is not running as PID 1, which causes the script to fail. As a result:  
- `dpkg` enters a broken state.  
- Any subsequent package installations or updates fail because `dpkg` keeps retrying the failing `postinstall` script.

**This is extremely critical for automated workflows.**

### **Error Output**  
Below is the error message observed during installation:

```text
Setting up otelcol-contrib (0.115.1) ...
Created symlink /etc/systemd/system/multi-user.target.wants/otelcol-contrib.service → /lib/systemd/system/otelcol-contrib.service.
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
dpkg: error processing package otelcol-contrib (--install):
 installed otelcol-contrib package post-installation script subprocess returned error exit status 1
Errors were encountered while processing:
 otelcol-contrib
```

### **Steps to Reproduce**  
1. Enter a `chroot` or `systemd-nspawn` environment.  
2. Install `otelcol` using `dpkg` (e.g., `dpkg -i otelcol-contrib.deb`).  
3. Observe that the installation fails with the above error message due to `systemctl` commands in `postinstall.sh`.  

### **Expected Behavior**  
The script should check whether `systemd` is running before executing `systemctl` commands. If `systemd` is not active, the script should safely skip service-related operations.

### **Proposed Solution**  
Modify `postinstall.sh` to verify if `systemd` is running as PID 1 before calling `systemctl`. For example:

```bash
if ps --no-headers -o comm 1 | grep -q 'systemd'; then
    # Safe to run systemctl commands
    systemctl start otelcol-contrib
fi
```

This ensures that the `systemctl` commands only run in appropriate environments, avoiding failures in chrooted setups.

mowies:
Maybe @netsandbox has some input here, since from past PRs he showed experience with these things 🤔 
@bowling233 are you willing to file a PR for this?

netsandbox:
Systemd RPM macros append ` || :` to all systemctl commands to not fail package scripts: https://github.com/systemd/systemd/blob/main/src/rpm/macros.systemd.in

[...]

PR #874, opened 2025-03-12
fix(postinstall): ignore systemctl commands when systemd is not running

bowling233:
Based on #767. This avoids package manager like `dpkg` to fail in chrooted setups.

### DIFF
--- a/.chloggen/fix_linux_package_scripts.yaml
+++ b/.chloggen/fix_linux_package_scripts.yaml
@@ -7,7 +7,7 @@ change_type: bug_fix
 component: packaging
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Fix Linux package scripts
+note: Ensure Linux package scripts work in environments without systemd such as chrooted environments
 
 # One or more tracking issues or pull requests related to the change
 issues: [767]
--- a/.chloggen/fix_linux_package_scripts.yaml
+++ b/.chloggen/fix_linux_package_scripts.yaml
@@ -1,0 +1,26 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: packaging
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix Linux package scripts
+
+# One or more tracking issues or pull requests related to the change
+issues: [767]
+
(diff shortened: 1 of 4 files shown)


mx-psi (commented):
Hi, can you add a changelog note? Also, could you expand on why this is needed in chrooted setups?

bowling233:
Hi, I've added a changelog.

When working in a chrooted environment, you're effectively isolating a portion of the filesystem and running commands within that confined space. However, systemd (the init system) is not running as PID 1 inside the chroot. This leads to issues when trying to use systemctl, as it relies on communicating with the systemd daemon, which is responsible for managing services and units.

```shell
$ systemctl daemon-reload
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
```

Many systemd-related commands, such as `systemctl daemon-reload` or `systemctl restart`, are not required when setting up a chrooted system.

In our scenario, we use a chroot environment in our CI/CD pipeline to build system images and install `otelcol`. However, during installation, `otelcol` executes `systemctl` commands, which fail in the chrooted environment.  

For package managers like dpkg, this leads to the following issues:  

- dpkg enters a broken state. 
- Subsequent package installations or updates fail because dpkg keeps retrying the failing post-installation script.  

Therefore, this fix would be very beneficial for CI/CD pipelines.

mx-psi:
Thanks for being patient with me and providing all this context :) 

> If the main concern is suppressing meaningful errors, we could add checks like [ -d /run/systemd/system ] or systemctl is-system-running --quiet to verify whether the systemd daemon is running before executing commands.

I think this is a reasonable approach. I understand and sympathize with distro maintainers not wanting to troubleshoot this every time for every package, but given that, for our particular case, this seems to work fine for most users, I would rather take the gentler approach of checking here. We can always go more aggressive with `|| :` later

bowling233:
The changes make it much clearer now—looks great. 🚀

.chloggen/fix_linux_package_scripts.yaml
```diff
@@ -0,0 +1,26 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: packaging
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix Linux package scripts
```
mx-psi, line 10:
```suggestion
note: Ensure Linux package scripts work in environments without systemd such as chrooted environments
```

mx-psi (approved):
Looks great, thanks! Will make a small change on the release note

PR #893, opened 2025-03-27
feat(otelcol-contrib): add GitLab receiver module to manifest

flc1125:
I'm not entirely sure about this modification. It's just that when I was using the GitLab receiver, I noticed it was missing this component. So I tried to find it here.

### DIFF
--- a/.chloggen/contrib-add-gitlabreceiver.yaml
+++ b/.chloggen/contrib-add-gitlabreceiver.yaml
@@ -4,14 +4,13 @@
 change_type: new_component
 
 # The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
-component: schemaprocessor
+component: gitlabreceiver
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: |
-  Add Schema Processor to contrib
+note: Add Gitlab Receiver to contrib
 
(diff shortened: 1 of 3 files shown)


atoulme:
It is still in development, so it's not ready to be added to contrib: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/gitlabreceiver

flc1125:
> It is still in development, so it's not ready to be added to contrib: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/gitlabreceiver

Thank you for your answer. I wish to use it, I have looked up some related information and found that it seems to be able to solve my problem: https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder

atoulme:
You can petition to move it to alpha, open an issue in contrib and make your case. It landed recently.

flc1125:
> You can petition to move it to alpha, open an issue in contrib and make your case. It landed recently.

It is a good method. However, in reality, I haven't delved deeply into using this component yet. I just discovered it, and I want to apply it. Let's see how the existing data looks.

Therefore, I am not yet able to determine if this component has reached the alpha stage. I also don't fully understand the author's expectations for this component (i.e., what the author envisions as alpha).

I need to try some practical applications before I can make a judgment.

PR #901, opened 2025-04-01
Revert "[chore] Add dockerhub metadata with links to readme files (#8…

TylerHelmuth:
…73)"

This reverts commit 101c9dd00ae52b35ddaf27fa6aadc2446d32d1d1.

This commit added the `dockerhub` section to the goreleaser config but didn't set a username, which broke the release process: https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/14197524938/job/39776413907

### DIFF
--- a/cmd/goreleaser/internal/configure.go
+++ b/cmd/goreleaser/internal/configure.go
@@ -41,17 +41,11 @@ const (
 )
 
 var (
-	baseArchs                 = []string{"386", "amd64", "arm", "arm64", "ppc64le", "s390x"}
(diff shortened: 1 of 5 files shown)


mowies (approved):

dehaansa (approved):

mowies:
I found my mistake for this: the`DOCKER_USERNAME` and `DOCKER_PASSWORD` are not set in the environment but only in the secrets, so those variables would need to be added.

Then the `otelbot` user would be used to update info on the dockerhub page. The question is if that user has enough privileges to do so. I don't think i can check that by myself.

PR #908, opened 2025-04-10
Introduce eBPF Profiler distribution

dmathieu:
This introduces a new distribution that is meant to run the [eBPF Profiler](https://github.com/open-telemetry/opentelemetry-ebpf-profiler) as an agent.


distributions/otelcol-ebpf-profiler/manifest.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+processors:
+  - gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.123.0
+  - gomod: go.opentelemetry.io/collector/processor/memorylimiterprocessor v0.123.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sattributesprocessor v0.123.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourcedetectionprocessor v0.123.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourceprocessor v0.123.0
+
+receivers:
+  - gomod: go.opentelemetry.io/ebpf-profiler v0.0.0-20250408174414-ead430a15d49
```
TylerHelmuth, line 23: This doesn't look like it comes from Core or Contrib, which breaks one of the criteria

distributions/otelcol-ebpf-profiler/README.md
```diff
@@ -0,0 +1,24 @@
...
+well as a subset of components from [OpenTelemetry Collector
+Core](https://github.com/open-telemetry/opentelemetry-collector) and
+[OpenTelemetry Collector
+Contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib).
+
+## Components
+
+The full list of components is available in the [manifest](manifest.yaml).
+
+### Rules for Component Inclusion
```
TylerHelmuth, line 17: Please add a rule for which processors and extensions are included.

distributions/otelcol-ebpf-profiler/manifest.yaml
```diff
@@ -0,0 +1,37 @@
+dist:
+  module: github.com/open-telemetry/opentelemetry-collector-releases/ebpf-profiler
+  name: otelcol-ebpf-profiler
+  description: OpenTelemetry Collector for eBPF Profiling
+  version: 0.123.1
+  output_path: ./_build
+
+exporters:
```
TylerHelmuth, line 8: The loadbalancing exporter speaks OTLP, should it be included?

distributions/otelcol-ebpf-profiler/manifest.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - gomod: go.opentelemetry.io/collector/confmap/provider/envprovider v1.29.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/fileprovider v1.29.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/httpprovider v1.29.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/httpsprovider v1.29.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/yamlprovider v1.29.0
+
+# When adding a replace, add a comment before it to document why it's needed and when it can be removed
+replaces:
+  # see https://github.com/openshift/api/pull/1515
+  - github.com/openshift/api => github.com/openshift/api v0.0.0-20230726162818-81f778f3b3ec
```
TylerHelmuth, line 37: The other distributions don't need this, what about this distribution does?

distributions/otelcol-ebpf-profiler/manifest.yaml
dmathieu: It does not seem to speak profiles

distributions/otelcol-ebpf-profiler/README.md
dmathieu: I've reworded things, I hope that's what you expected.

distributions/otelcol-ebpf-profiler/manifest.yaml
dmathieu: So doe the module itself.
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/go.mod#L192

Every time I've used this module so far, I needed to include this replace. I've been a bit surprised by that, but I assumed this was a known issue.

.github/workflows/release-ebpf-profiler.yaml
```diff
@@ -0,0 +1,16 @@
...
+    tags: ["v*"]
+
+jobs:
+  release:
+    name: Release eBPF Profiler
+    uses: ./.github/workflows/base-release.yaml
+    with:
+      distribution: otelcol-ebpf-profiler
+      goos: '[ "linux" ]'
+      goarch: '[ "amd64", "arm64", "ppc64le", "s390x" ]'
```
christos68k, line 14: The profiler receiver doesn't work on `ppc64le` and `s390x` architectures.

```suggestion
      goarch: '[ "amd64", "arm64" ]'
```

.github/workflows/ci-goreleaser-ebpf-profiler.yaml
```diff
@@ -0,0 +1,37 @@
...
+  check-goreleaser:
+    name: CI - eBPF Profiler - GoReleaser
+    uses: ./.github/workflows/base-ci-goreleaser.yaml
+    with:
+      distribution: otelcol-ebpf-profiler
+      config_file: ebpf-profiler-config
+      docker_run_options: '--privileged -v /sys/kernel/debug/:/sys/kernel/debug/'
+      otelcol_run_options: '--feature-gates=service.profilesSupport'
+      goos: '[ "linux" ]'
+      goarch: '[ "amd64" ]'
```
christos68k, line 36: Does it make sense to add `arm64` here?

cmd/goreleaser/internal/configure.go
```diff
@@ -45,6 +46,7 @@ var (
 	winContainerArchs = []string{"amd64"}
 	darwinArchs       = []string{"amd64", "arm64"}
 	k8sArchs          = []string{"amd64", "arm64", "ppc64le", "s390x"}
+	ebpfProfilerArchs = []string{"amd64"}
```
christos68k, line 49: Does it make sense to add `arm64` here?

```suggestion
	ebpfProfilerArchs = []string{"amd64", "arm64"}
```

cmd/goreleaser/internal/configure.go
```diff
@@ -141,6 +143,20 @@ var (
 			newContainerImageManifests(d.name, "linux", k8sArchs, containerImageOptions{}),
 		)
 	}).WithDefaultArchives().WithDefaultChecksum().WithDefaultSigns().WithDefaultDockerSigns().WithDefaultSBOMs().Build()
+
+	// ebpf-profiler distro
+	ebpfProfilerDist = newDistributionBuilder(ebpfProfilerDistro).WithConfigFunc(func(d *distribution) {
+		d.buildConfigs = []buildConfig{
+			&fullBuildConfig{targetOS: "linux", targetArch: ebpfProfilerArchs, ppc64Version: []string{"power8"}},
```
christos68k, line 150: Do we need `ppc64Version` here?

```suggestion
			&fullBuildConfig{targetOS: "linux", targetArch: ebpfProfilerArchs},
```

distributions/otelcol-ebpf-profiler/.goreleaser.yaml
```diff
@@ -0,0 +1,79 @@
...
+release:
+  replace_existing_artifacts: true
+builds:
+  - id: otelcol-ebpf-profiler-linux
+    goos:
+      - linux
+    goarch:
+      - amd64
+    goppc64:
+      - power8
```
christos68k, line 16: Do we need this?

```suggestion
```

christos68k (approved):
LGTM thanks!

distributions/otelcol-ebpf-profiler/.goreleaser.yaml
dmathieu: This is auto-generated ;)

cmd/goreleaser/internal/configure.go
```diff
@@ -384,6 +400,7 @@ type distribution struct {
 	dockerSigns             []config.Sign
 	sboms                   []config.SBOM
 	checksum                config.Checksum
+	enableCgo               bool
```
mx-psi, line 403: Can we move this `enableCgo` changes to a separate PR that we can land more quickly?

distributions/otelcol-ebpf-profiler/manifest.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.123.0
+  - gomod: go.opentelemetry.io/collector/processor/memorylimiterprocessor v0.123.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sattributesprocessor v0.123.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourcedetectionprocessor v0.123.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourceprocessor v0.123.0
+
+receivers:
+  - gomod: go.opentelemetry.io/ebpf-profiler v0.0.0-20250408174414-ead430a15d49
+    import: go.opentelemetry.io/ebpf-profiler/collector
+    name: profilesreceiver
```
mx-psi, line 25: Not a blocker for this PR but it would be less confusing for users if the import and the name matched and they can use this component like any other

cmd/goreleaser/internal/configure.go
dmathieu: Sure: https://github.com/open-telemetry/opentelemetry-collector-releases/pull/913

distributions/otelcol-ebpf-profiler/manifest.yaml
dmathieu: Actually, we don't need the name. I've removed it.

mx-psi (commented):
This distribution needs cgo, which means a dependency on libc (and, specifically glibc). What are our expectations regarding support for older glibc versions (and thus older OS versions)? Could we use musl instead?

christos68k:
> This distribution needs cgo, which means a dependency on libc (and, specifically glibc). What are our expectations regarding support for older glibc versions (and thus older OS versions)? Could we use musl instead?

We're using `musl` inside Elastic, we just haven't propagated it to the OTel [repo](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pulls) yet. We know it works in our standalone agent build, my only concern would be around the build system and cross-compilation, but pretty sure we can make it happen if we use a Docker 'builder' image.

.github/workflows/ci-goreleaser-ebpf-profiler.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - "go.sum"
+
+jobs:
+  check-goreleaser:
+    name: CI - eBPF Profiler - GoReleaser
+    uses: ./.github/workflows/base-ci-goreleaser.yaml
+    with:
+      distribution: otelcol-ebpf-profiler
+      config_file: ebpf-profiler-config
+      docker_run_options: '--privileged -v /sys/kernel/debug/:/sys/kernel/debug/'
```
florianl, line 33: The eBPF profiler needs to get access to the host pid and it does not require write access mapping to /sys/kernel/debug
```suggestion
      docker_run_options: '--privileged --pid=host -v /sys/kernel/debug/:/sys/kernel/debug/:ro'
```

cmd/goreleaser/internal/configure.go
```diff
@@ -392,18 +409,22 @@ func (d *distribution) BuildProject() config.Project {
 		builds = append(builds, buildConfig.Build(d.name))
 	}
 
+	env := []string{
+		"COSIGN_YES=true",
+		"LD_FLAGS=-s -w",
+		"BUILD_FLAGS=-trimpath",
+	}
+	if !d.enableCgo {
```
florianl, line 417: Should this print out a warning or return an error, if the project is `ebpfProfilerDist`?

distributions/otelcol-ebpf-profiler/.goreleaser.yaml
```diff
@@ -0,0 +1,96 @@
...
+    binary: otelcol-ebpf-profiler
+    ldflags:
+      - '{{ .Env.LD_FLAGS }}'
+    flags:
+      - '{{ .Env.BUILD_FLAGS }}'
+archives:
+  - id: otelcol-ebpf-profiler
+    builds:
+      - otelcol-ebpf-profiler-linux
+    name_template: '{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}{{ if .Mips }}_{{ .Mips }}{{ end }}'
```
florianl, line 26: Mips support can be dropped, I think.
```suggestion
    name_template: '{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}{{ end }}'
```

distributions/otelcol-ebpf-profiler/.goreleaser.yaml
```diff
@@ -0,0 +1,96 @@
+version: 2
+project_name: opentelemetry-collector-releases
+env:
+  - COSIGN_YES=true
+  - LD_FLAGS=-s -w
```
florianl, line 5: The eBPF Profiler uses some more LD flags - would this be the correct place to add them as well?
https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/ead430a15d497736841ee58436ed4f9a6dc86703/Makefile#L43-L46

mx-psi:
> > This distribution needs cgo, which means a dependency on libc (and, specifically glibc). What are our expectations regarding support for older glibc versions (and thus older OS versions)? Could we use musl instead?
> 
> We're using `musl` inside Elastic, we just haven't propagated it to the OTel [repo](https://github.com/open-telemetry/opentelemetry-ebpf-profiler) yet. We know it works in our standalone agent build, my only concern would be around the Collector build system and cross-compilation, but pretty sure we can make it happen if we use a Docker 'builder' image.
> 
> Another reason for using `musl` is releasing static binaries. Static linking with glibc is problematic on multiple fronts.

Awesome to hear that `musl` is working well! I think this is a solvable problem upstream as well. My main concern until we use musl upstream is to clearly document what glibc versions we support. Could we add something to the distribution's README?

rockdaboot:
> > > This distribution needs cgo, which means a dependency on libc (and, specifically glibc). What are our expectations regarding support for older glibc versions (and thus older OS versions)? Could we use musl instead?
> > 
> > 
> > We're using `musl` inside Elastic, we just haven't propagated it to the OTel [repo](https://github.com/open-telemetry/opentelemetry-ebpf-profiler) yet. We know it works in our standalone agent build, my only concern would be around the Collector build system and cross-compilation, but pretty sure we can make it happen if we use a Docker 'builder' image.
> > Another reason for using `musl` is releasing static binaries. Static linking with glibc is problematic on multiple fronts.
> 
> Awesome to hear that `musl` is working well! I think this is a solvable problem upstream as well. My main concern until we use musl upstream is to clearly document what glibc versions we support. Could we add something to the distribution's README?

Using musl also has downsides: No support of NSS or mDNS, requires static host/ip table to be /etc/hosts.
Static linking of glibc: It is not 100% static, thus possibly missing symbols when running the app on older environments and name resolution has potential issues (NSS). There might be other issues as well.

Ideally we have muslc static linking in combination with a separate resolver library.

rockdaboot (approved):

.github/workflows/ci-goreleaser-ebpf-profiler.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - "go.mod"
+      - "go.sum"
+
+jobs:
+  check-goreleaser:
+    name: CI - eBPF Profiler - GoReleaser
+    uses: ./.github/workflows/base-ci-goreleaser.yaml
+    with:
+      distribution: otelcol-ebpf-profiler
+      config_file: ebpf-profiler-config
```
mowies, line 32: why leave out the file extention here? i immediately confused this with a folder name. maybe it would be better for readability to include the extension.

```suggestion
      config_file: ebpf-profiler-config.yaml
```

(needs to be adjusted in the `base-ci-goreleaser.yaml` file as well

cmd/goreleaser/internal/configure.go
dmathieu: Hmm I'm not sure I see the point. Checking for project name here kind of defeats the purpose of having an option.
In that case, let's not have this config, and use a switch statement with project names (I don't think that'd be a good idea).

Maybe I don't understand what you want though.

.github/workflows/ci-goreleaser-ebpf-profiler.yaml
dmathieu: I wanted to avoid config be something else than `yaml`. But sure, I can change this.

distributions/otelcol-ebpf-profiler/.goreleaser.yaml
dmathieu: Done.
Not that the vc flags will have similar data added by goreleaser.
https://goreleaser.com/cookbooks/using-main.version/

.github/workflows/release-ebpf-profiler.yaml
```diff
@@ -0,0 +1,16 @@
...
+    tags: ["v*"]
+
+jobs:
+  release:
+    name: Release eBPF Profiler
+    uses: ./.github/workflows/base-release.yaml
+    with:
+      distribution: otelcol-ebpf-profiler
+      goos: '[ "linux" ]'
+      goarch: '[ "amd64" ]'
```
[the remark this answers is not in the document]
florianl: Just found https://github.com/open-telemetry/opentelemetry-collector-releases/pull/908/files#r2043906410

distributions/otelcol-ebpf-profiler/.goreleaser.yaml
```diff
@@ -0,0 +1,97 @@
+version: 2
+project_name: opentelemetry-collector-releases
+env:
+  - COSIGN_YES=true
+  - LD_FLAGS=-extldflags=-static
+  - BUILD_FLAGS=-trimpath
+  - GO_TAGS=osusergo,netgo
```
florianl, line 7: Should we add `CGO_ENABLED=1` here?
```suggestion
  - GO_TAGS=osusergo,netgo
  - CGO_ENABLED=1
```
`CGO_ENABLED=1` is required for the project.

cmd/goreleaser/internal/configure.go
```diff
@@ -141,6 +143,22 @@ var (
...
 		)
 	}).WithDefaultArchives().WithDefaultChecksum().WithDefaultSigns().WithDefaultDockerSigns().WithDefaultSBOMs().Build()
+
+	// ebpf-profiler distro
+	ebpfProfilerDist = newDistributionBuilder(ebpfProfilerDistro).WithConfigFunc(func(d *distribution) {
+		d.buildConfigs = []buildConfig{
+			&fullBuildConfig{targetOS: "linux", targetArch: ebpfProfilerArchs},
+		}
+		d.containerImages = slices.Concat(
+			newContainerImages(d.name, "linux", ebpfProfilerArchs, containerImageOptions{armVersion: "7"}),
```
florianl, line 153: [ARM7 usually](https://en.wikipedia.org/wiki/ARM_architecture_family#Cores) referees to 32 bit architectures. While ARM8 and the following support 64 bit widths.
If `"7"` referees here to the regular ARM architecture, then it should be bumped to `"8"`, as eBPF profiler is limited to 64 bit architectures.

dmathieu: We don't actually need this option at all, as it would only be used for the `arm` arch. We only use `arm64`.
I have removed it.

florianl (approved):

distributions/otelcol-ebpf-profiler/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM golang:1.23.8
```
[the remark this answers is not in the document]
dmathieu: Running on scratch is missing the linked libraries we need to run.
See this CI run which was triggered in the early stages of this PR, when it relied on scratch: https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/14381773455/job/40327364431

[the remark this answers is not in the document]
mx-psi: Is there a way we can pull just the libraries we need instead of all of `golang:1.23.8`?

.github/workflows/ci-goreleaser-ebpf-profiler.yaml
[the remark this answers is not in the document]
dmathieu: Sure. I have removed `.github/workflows/release-ebpf-profiler.yaml`.

distributions/otelcol-ebpf-profiler/Dockerfile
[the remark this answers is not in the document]
dmathieu: Possibly, but manually picking the dlopen, dlfree, (m)alloc and free operations seems brittle.
Any regression willn also be hard to investigate, as the errors aren't really explanatory (the binary is just marked as not found when any linked library isn't found).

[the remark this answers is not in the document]
mx-psi: Fair enough, I still would like to investigate eventually using a more lean image, but we can move this discussion into an issue

mx-psi (approved):
Only thing missing is documenting what glibc versions this works with, otherwise it's good to go from my side

cmd/goreleaser/internal/configure.go
```diff
@@ -45,6 +46,7 @@ var (
 	winContainerArchs = []string{"amd64"}
 	darwinArchs       = []string{"amd64", "arm64"}
 	k8sArchs          = []string{"amd64", "arm64", "ppc64le", "s390x"}
+	ebpfProfilerArchs = []string{"amd64", "arm64"}
```
andrzej-stencel, line 49: I'm probably missing something; why are we building the eBPF distro for ARM64 here if it doesn't work according to the [comment above](https://github.com/open-telemetry/opentelemetry-collector-releases/pull/908/files#r2038983667)?

distributions/otelcol-ebpf-profiler/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+- Only includes components from Contrib and Core, except for the ebpf-profiler receiver.
+- Only components that are Alpha or higher.
+- All processors, connectors, and extensions must facilitate the collection and processing of data that is generated by the eBPF profiler.
```
andrzej-stencel, line 30: It's interesting to see that there are currently no connectors that support profies.

dmathieu: That's going to change soon. I have a PR adding support for the count connector.

cmd/goreleaser/internal/configure.go
dmathieu: Good point. It's added to goreleaser, but not used by the action afterward. I've removed it.

dmathieu:
> Only thing missing is documenting what glibc versions this works with, otherwise it's good to go from my side

It doesn't look like the ebpf-profiler documents that itself, so I'm not really sure how to do document that here.

mx-psi:
> > Only thing missing is documenting what glibc versions this works with, otherwise it's good to go from my side
> 
> It doesn't look like the ebpf-profiler documents that itself, so I'm not really sure how to do document that here.

The glibc version in the 24.04 runner would be 2.39:

```
❯ docker run --rm -it ubuntu:24.04 ldd --version
ldd (Ubuntu GLIBC 2.39-0ubuntu8.4) 2.39
Copyright (C) 2024 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
```

I think that's the minimum version we can guarantee

andrzej-stencel (approved):
Thank you @dmathieu, looks good from my side.

dmathieu:
I've added a mention of the current glibc version.

mx-psi:
cc @open-telemetry/collector-contrib-maintainers This adds a new distribution for profiling. It does not add it to the release process yet (no artifact will be available for end-users yet). Instead, it only tests that the distribution can be built. Since this does not have end-user impact, I think we can merge this PR by EOW unless there are blocking comments

distributions/otelcol-ebpf-profiler/manifest.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - gomod: go.opentelemetry.io/collector/exporter/otlpexporter v0.123.0
+  - gomod: go.opentelemetry.io/collector/exporter/otlphttpexporter v0.123.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/fileexporter v0.123.0
+
+processors:
+  - gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.123.0
+  - gomod: go.opentelemetry.io/collector/processor/memorylimiterprocessor v0.123.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sattributesprocessor v0.123.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourcedetectionprocessor v0.123.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourceprocessor v0.123.0
```
songy23, line 20: FYI These versions are not up-to-date and are not auto-updated by the releaser https://github.com/open-telemetry/opentelemetry-collector-releases/pull/933/commits/aa933b1749202d41d00f75830310a3e1945298f2

dmathieu: Hmm this is strange. `bump-versions.sh` has this manifest configured, and everything in collector-contrib should be upgraded from there.

PR #912, opened 2025-04-11
feat: add cgroup runtime extension to contrib and k8s manifest

rogercoll:


swiatekm (changes requested):
This extension has stability "development", but both the contrib and k8s distributions require at least Alpha.

mowies (approved):

swiatekm (approved):

PR #913, opened 2025-04-11
Add option to allow enabling cgo in distributions

dmathieu:
This will be needed by the eBPF profiler distributions.


dmathieu:
I don't think this should need a changelog entry, as it's an internal option which isn't being used yet.

andrzej-stencel (approved):

mx-psi (approved):

PR #919, opened 2025-04-15
Add support for custom ldtags and go tags

dmathieu:
So the eBPF profiler can use custom ld tags, as well as Go tags.

See https://github.com/open-telemetry/opentelemetry-collector-releases/pull/908#discussion_r2041460234


dmathieu:
This shouldn't need a changelog entry.

mx-psi (approved):

PR #944, opened 2025-05-08
[distributions/otlp] add more component to the otlp-only distribution

### LINKED ISSUE #669: Add more components to OTLP distro
As discussed in https://github.com/open-telemetry/opentelemetry-collector-releases/pull/663#discussion_r1765134837, we should consider adding the following components to the OTLP distro eventually:

- [ ] auth extension(s)
- [ ] pprof extension
- [ ] debug exporter
- [ ] healthcheck v2 extension

cuichenli:
fixes #669

### DIFF
--- a/distributions/otelcol-otlp/manifest.yaml
+++ b/distributions/otelcol-otlp/manifest.yaml
@@ -6,11 +6,25 @@ dist:
   output_path: ./_build
 
 exporters:
+  - gomod: go.opentelemetry.io/collector/exporter/debugexporter v0.125.0
   - gomod: go.opentelemetry.io/collector/exporter/otlpexporter v0.125.0
   - gomod: go.opentelemetry.io/collector/exporter/otlphttpexporter v0.125.0
 
+
+extensions:
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/basicauthextension v0.125.0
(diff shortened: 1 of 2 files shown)


mowies (approved):

distributions/otelcol-otlp/manifest.yaml
```diff
@@ -6,12 +6,26 @@ dist:
...
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/bearertokenauthextension v0.125.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/healthcheckextension v0.125.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/pprofextension v0.125.0
+
+
 receivers:
   - gomod: go.opentelemetry.io/collector/receiver/otlpreceiver v0.125.0
 
+processors:
+  - gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.125.0
```
mx-psi, line 25: We should definitely not add this one, we are still discussing whether this is the right call for this distribution or not

mx-psi (changes requested):
Sorry that I didn't get to replying sooner. I think this needs a bit more discussion. I am not sure if we want to add these until we have marked them as 1.x.

If you are interested in working on marking these as 1.x, I think the auth extensions may be doable, but please speak to me before starting any work, this would need coordination

PR #982, opened 2025-06-17
[chore] Use DOCKER_TOKEN_COLLECTOR_RELEASES rather than DOCKER_PASSWORD

dmathieu:
We're going away from using a an org-wide token to use project-scoped credentials.
So this switches the Docker token from the global one to a repo-specific one (which has been set already).

See https://github.com/open-telemetry/community/issues/2805

Every distribution should already have access. But please review the mentioned issue to ensure I haven't forgotten anything.


mowies (changes requested):
Seems like the `otel/opentelemetry-collector-opampsupervisor` is missing from the list in the original issue. I commented there to add it as well.

mowies (approved):

songy23 (approved):

codeboten (approved):

Issue #653, opened 2024-08-28
StatsD Receiver in the K8S distrib


maxime-menard:
We can't use the OpenTelemetry Collector Kubernetes Distro because StatsD receiver is not included.
We can obviously fallback to the contrib one, but this is much bigger. Would it be ok to include the StatsD receiver in the k8s one ?

PR #1015, opened 2025-07-15
feat(distribution/otelcol-k8s): introduce statsdreceiver

clavinjune:
Adding StatsD Receiver to otelcol-k8s

solves #653


swiatekm (changes requested):
Adding this component would break the following inclusion rule:

> - All receivers must facilitate the collection of data that is generated by Kubernetes or services running in Kubernetes

clavinjune:
got it @swiatekm, should we close this PR, then?

PR #1016, opened 2025-07-16
[chore][WIP] add nightly release setup with pipeline and goreleaser settings

mowies:


.github/workflows/base-release.yaml
```diff
@@ -16,13 +16,19 @@ on:
...
+        default: false
 
 env:
   # renovate: datasource=github-tags depName=goreleaser-pro packageName=goreleaser/goreleaser-pro
   GORELEASER_PRO_VERSION: v2.7.0
 
 jobs:
   prepare:
+    outputs:
+      GORELEASER_PREVIOUS_TAG: ${{ steps.prev-tag.outputs.GORELEASER_PREVIOUS_TAG }}
```
mx-psi: nit: it's a bit unclear what this is, at a first glance it may seem to be "the last released tag of goreleaser" instead of "the last tag released on this repository"

distributions/otelcol-contrib/.goreleaser.yaml
```diff
@@ -4,6 +4,7 @@ env:
   - COSIGN_YES=true
   - LD_FLAGS=-s -w
   - BUILD_FLAGS=-trimpath
+  - CONTAINER_IMAGE_EPHEMERAL_TAG={{ if .IsNightly }}nightly{{ else }}latest{{ end }}
```
mx-psi, line 7: I think we could do a first change that just defines this var and sets it unconditionally to `latest` (sorry, I am a bit scared about changes in this repo and want to have the cleanest diff possible)

mowies: that's reasonable :)
i'll create a separate PR

mowies: https://github.com/open-telemetry/opentelemetry-collector-releases/pull/1036

.github/workflows/base-release.yaml
mowies: i'll try using some better variables throughout the pipeline, but goreleaser still needs that name in the end :)
[source](https://goreleaser.com/cookbooks/set-a-custom-git-tag/)

mx-psi: Ah, bummer!

cmd/builder/.goreleaser.yml
```diff
@@ -103,9 +103,6 @@ docker_manifests:
       - ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-builder:{{ .Env.CONTAINER_IMAGE_EPHEMERAL_TAG }}-ppc64le
 release:
   make_latest: false
-  github:
-    owner: open-telemetry
-    name: opentelemetry-collector-releases
```
mx-psi: Why is this removed?

.github/workflows/base-release.yaml
mowies: i did change the variable throughout the whole pipeline now and only referenced the `GORELEASER_PREVIOUS_TAG` name where it's actually needed, in the goreleaser action

cmd/builder/.goreleaser.yml
mowies: Goreleaser always takes the current repo as default anyways. This setting would usually be used to release code into a different repo, which we don't do.

cmd/builder/.goreleaser.yml
```diff
@@ -141,4 +138,8 @@ sboms:
     artifacts: archive
   - id: package
     artifacts: package
-    
+nightly:
+  version_template: '{{ incpatch .Version}}-nightly.{{ .Now.Format "200601021504" }}'
+  tag_name: nightly-builder
+  publish_release: true
```
mx-psi, line 147: I wonder if this is going to be too spammy for people and we should set it to 'false'. Would we lose any benefits if we set this to false?

cmd/builder/.goreleaser.yml
mx-psi: I would rather do this change in a separate PR :sweat_smile: I think this is something we can merge before the next release, and then the nightlies PR can be merged right after the next release

cmd/builder/.goreleaser.yml
mowies: I'm not sure how the notifications work on github for this. The plan would be to have a rolling release with rolling tags that will be updated instead of creating new ones.
If we don't publish releases, it's hard to impossible to get an asset from a nightly release, since those are attached to the github releases usually. I think it's still possible to fetch them through the GH API somehow even without a published release but I didn't manage to do that successfully until now.

mx-psi: > If we don't publish releases, it's hard to impossible to get an asset from a nightly release, since those are attached to the github releases usually.

So, to confirm, if we did something like draft releases, we would not be able to get the assets?

mowies: there's no option for drafts, but i'll try an see what happens when i turn this setting of just to be sure

mx-psi: Oki, I think drafts would be ideal if we can get the assets, since only people with write permissions would be able to check them

cmd/builder/.goreleaser.yml
mowies: sure, i'll move that change out

mowies: https://github.com/open-telemetry/opentelemetry-collector-releases/pull/1041

.github/workflows/base-release.yaml
```diff
@@ -70,8 +76,25 @@ jobs:
           go-version: "~1.24"
           check-latest: true
 
+      - name: Set goreleaser last tag reference in case of non-nightly release
+        id: prev-tag
+        if: inputs.nightly != 'true'
+        # find previous tag by filtering out nightly tags and binary release (OCB, OpAMP) tags and then choosing the
+        # second to last tag (last one is the current release)
```
douglascamata, line 85: By current mean the release that is currently being done, right?

.github/workflows/base-release.yaml
```diff
@@ -141,7 +165,7 @@ jobs:
           if-no-files-found: error
 
   release:
-    name: ${{ inputs.distribution }} Release
+    name: Release
```
douglascamata, line 170: Don't we want to keep the distribution in the name anymore? I think it would even be nice to indicate whether it was a nightly release. 🤔

.github/workflows/nightly-release.yaml
```diff
@@ -0,0 +1,43 @@
+name: Nightly Release
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 2 * * *" # release nightly at 2AM
```
douglascamata, line 6: Curious question: do you know what's the timezone of the servers that are running this?

.github/workflows/nightly-release.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+permissions:
+  contents: read
+
+env:
+  # renovate: datasource=github-tags depName=goreleaser-pro packageName=goreleaser/goreleaser-pro
+  GORELEASER_PRO_VERSION: v2.7.0
```
douglascamata:
```suggestion
  GORELEASER_PRO_VERSION: v2.11.0
```

.github/workflows/nightly-release.yaml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.GH_PAT }}
```
douglascamata: In the base release workflow this secret has been renamed from `GH_PAT` to `GITHUB_TOKEN`. Why was it renamed there and here we used the old name?

.github/workflows/nightly-release.yaml
```diff
@@ -0,0 +1,43 @@
...
+          git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+
+          # A previous release was created using a lightweight tag
+          # git describe by default includes only annotated tags
+          # git describe --tags includes lightweight tags as well
+          DESCRIBE=`git tag -l --sort=-v:refname | grep -v cmd | grep -v nightly | head -n 1`
+          MAJOR_VERSION=`echo $DESCRIBE | awk '{split($0,a,"."); print a[1]}'`
+          MINOR_VERSION=`echo $DESCRIBE | awk '{split($0,a,"."); print a[2]}'`
+          MINOR_VERSION="$((${MINOR_VERSION} + 1))"
+          TAG="${MAJOR_VERSION}.${MINOR_VERSION}.0-nightly.$(date +'%Y%m%d%H%M')"
```
douglascamata: Maybe to help people understand what's happening here you could add in the comments what ends up being the contents of `$DESCRIBE`. This makes it easier to verify that the `awk` commands below are correct.

Curious question: why the patch version stays at zero?

.github/workflows/release-opampsupervisor.yaml
```diff
@@ -22,23 +22,51 @@ jobs:
...
+          echo "NIGHTLY_FLAG=--nightly" >> "$GITHUB_OUTPUT"
+
+      - name: Set COLLECTOR_REF
+        id: collector-ref
+        run: |
+          if [ '${{ contains(github.ref, '-nightly') }}' == 'true' ]; then
+            echo "COLLECTOR_REF=main" >> "$GITHUB_OUTPUT"
+          else
+            echo "COLLECTOR_TAG=${{ github.ref }}" >> "$GITHUB_OUTPUT"
+          fi
```
douglascamata, line 48: I noticed that these 3 steps are repeated in a few places. Could we somehow extract them into something reusable? Not required for now, but in the near term future it would be useful so that fixes and changes could easily be applied in all places.

douglascamata (commented):
Awesome work. This will be so helpful!

I left some minor comments/suggestions.

.github/workflows/nightly-release.yaml
mx-psi: Time is expressed in UTC https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule

.github/workflows/nightly-release.yaml
mx-psi: Can we make this change separately?

douglascamata: You mean the uprade, @mx-psi? Or the whole file? The version I suggested matches all the other files. It would be weird to have the nightlies running on an older version.

.github/workflows/base-release.yaml
mowies: that didn't work and it literally showed the curly braces statement in titles, that's why i removed it 😅

.github/workflows/base-release.yaml
mowies: exactly

.github/workflows/release-opampsupervisor.yaml
mowies: sounds good, that should be possible with another [reusable workflow](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows)

mowies: i'll put a note into the follow ups in the description. i'll probably create new issue from that after the PR is merged.

.github/workflows/nightly-release.yaml
mowies: In the base-release it's fine to use the standard token and just give it proper permissions to be able to write content into the GH release for example.
But the nightly release needs the ability to push a tag that ALSO triggers other CI runs, which the standard GITHUB_TOKEN is not capable of. You need some other PAT or so to also trigger pipeline runs.

Here, the use case is that the nightly tag is pushed by one pipeline, which then triggers the release pipelines.

mowies: But I think for extra safety in this PR, i'll revert the change in the base release pipeline or put it into a separate PR

mowies: i separated the change out into https://github.com/open-telemetry/opentelemetry-collector-releases/pull/1043

.github/workflows/nightly-release.yaml
mowies: sure, i'll add some comments.
i chose to bump the minor version since that's what usually happens in the collector 🤷🏼

mx-psi:
Unfortunately I do get notified for every release, including pre-releases
image

douglascamata:
I think the only way to avoid the notification spam would be to push the nightly releases to another repository, which could be a good idea. We can keep all the code and official releases in this repository, just push the nightlies to a `-nightly` repo. 

What do you think about this idea, @mx-psi and @mowies?

@mowies, do you think this would be possible?

mx-psi:
Yeah either that or drafts would work

mowies:
> I think the only way to avoid the notification spam would be to push the nightly releases to another repository, which could be a good idea. We can keep all the code and official releases in this repository, just push the nightlies to a `-nightly` repo.

that is gonna be hard to implement in the goreleaser configs. i don't think you can set a different repo just for nightly releases

cmd/builder/.goreleaser.yml
mowies: i tried disabling `publish_release` but then no releases are created at all. Also no drafts unfortunately :(

mowies:
Since drafts are not created when `publish_release` is set to false, I think we should just go with not publishing releases at first and just check that the pipelines work ok. Maybe creating issues when they fail or so.

douglascamata:
They will need to add support for that `release.draft` setting also in the `nightly` section.

.github/workflows/nightly-release.yaml
mx-psi, line 1: I am not sure who gets notified when this fails, I guess having a 'create issue' step like we have for some workflows in contrib would be good here

mowies: still working on that :)

mowies: done ✔️

mowies: it's admittedly not very smart right now, but at least it's not spammy 😅 
example: https://github.com/mowies/opentelemetry-collector-releases/issues/18

mx-psi (approved):
LGTM cc @open-telemetry/collector-releases-approvers

PR #1038, opened 2025-07-24
[chore] Generate OCB and OpAMP goreleaser configs from code

### LINKED ISSUE #1035: Generate OCB and OpAMP goreleaser configs from code
Similar to what is done for all Collector distros, the OCB and OpAMP supervisor goreleaser configs should also be generated from code so that they are more maintainable and easier to adapt.

mowies:
Fixes #1035


cmd/goreleaser/internal/configure.go
```diff
@@ -344,14 +418,148 @@ func (b *distributionBuilder) WithDefaultChecksum() *distributionBuilder {
...
+		dir = ".core/cmd/builder"
+	case opampBinary:
+		dir = ".contrib/cmd/opampsupervisor"
+	}
+
+	return config.Monorepo{
+		TagPrefix: fmt.Sprintf("cmd/%s/", b.dist.name),
+		Dir:       dir,
+	}
+}
```
douglascamata: Maybe at this point we could introduce a more generic `func (b *distributionBuilder) WithBinaryMonorepo(dir string)` that knows how to configure the monorepo for the binaries? 

I would like to avoid having split logic inside the builders depending on the distribution and instead make it clear in the distribution's definition what's different in each one of them. The builders shouldn't care about which distribution they are building.

The same logic would apply to the release customization functions that are defined a few lines below, but with the signature being something like `func (b *distributionBuilder) WithDefaultBinaryRelease(headerContent string)`.

What do you think, @mowies?

mowies: sure thing, i updated it now to have the distro/binary specific things at the top together with the distro definitions

atoulme (approved):

evan-bradley (approved):
Thanks @mowies. Creating these in code is a huge improvement over YAML.

PR #1051, opened 2025-07-30
Revert "[chore] remove redundant goreleaser settings (#1041)"

tomatopunk:
#1041 should be reverted because it is blocking the collector snapshot build, as seen in #13493, #13494, and #13296. 
It does not work as expected because we do not trigger releases from opentelemetry-collector-releases; 

instead, releases are triggered in open-telemetry-collector. In that workflow, the release repository is checked out, and only if release.github.name is empty will the current repository’s .git be used.


tomatopunk:
ci log is there: https://github.com/tomatopunk/opentelemetry-collector/actions/runs/16616910118/job/47011569500

mowies:
oh, i didn't know that we have this backwards dependency to the core repo

mowies (approved):

PR #1056, opened 2025-07-30
[otelcol][otelcol-contrib] opencensus

atoulme:
Remove opencensus exporter and receiver from otelcol and otelcol-contrib distributions


mowies (changes requested):
please also remove the corresponding opencensus port from dockerfiles

atoulme:
thanks @mowies removed.

Issue #984, opened 2025-06-20
otelcol-ebpf-profiler image


sbquinlan:
Where can I find the image for the ebpf profiler? https://github.com/open-telemetry/opentelemetry-collector-releases/pull/958 states that it started releasing the image, but I can't find any an release asset or package associated with the profiler.

dmathieu:
That commit had to be reverted: https://github.com/open-telemetry/opentelemetry-collector-releases/pull/973
We should bring it back soon. In the mean time, you need to build the image yourself.

douglascamata:
It seems like we can enable this again? Any reason for not opening a PR out of your work on https://github.com/dmathieu/opentelemetry-collector-releases/commit/d3f291f92c8099bba7751ca4cba83fe8f34308f7, @dmathieu? 

I can take care of this, if you are busy with other things.

PR #1098, opened 2025-08-08
[feat] Configure release of the `ebpf-profiler` distro

douglascamata:
This should solve https://github.com/open-telemetry/opentelemetry-collector-releases/issues/984.

It should be ready, considering that the following issues are solved:

- https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/513
- https://github.com/open-telemetry/community/issues/2805

The Docker Hub image repository is ready: https://hub.docker.com/r/otel/opentelemetry-collector-ebpf-profiler.

### DIFF
--- a/.chloggen/main.yaml
+++ b/.chloggen/main.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: contrib
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: This feature introduces anomaly detection using the Online Isolation Forest algorithm applied to streaming - traces, metrics, or logs
+
+# One or more tracking issues or pull requests related to the change
+issues: [1124]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: 
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
--- a/.chloggen/sapmreceiver.yaml
+++ b/.chloggen/sapmreceiver.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: sapmreceiver
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove the sapmreceiver from contrib as deprecation period has passed.
+
+# One or more tracking issues or pull requests related to the change
+issues: [1125]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
--- a/.github/workflows/base-binary-release.yaml
+++ b/.github/workflows/base-binary-release.yaml
@@ -36,30 +36,13 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set goreleaser last tag reference in case of non-nightly release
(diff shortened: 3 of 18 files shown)


.github/workflows/base-package-tests.yaml
```diff
@@ -14,11 +14,15 @@ on:
         type: boolean
         default: false
         description: "Set to true if a GH issue should be generated upon failure"
+      runner_os:
+        required: false
+        type: string
+        default: ubuntu-22.04
```
douglascamata: Wanted to bump this to `ubuntu-24.04`, but the upgrade triggers errors in the package tests. These errors seem to be related to systemd. Sadly I don't have enough knowledge to debug and fix them. :/

.github/workflows/ci-goreleaser-ebpf-profiler.yaml
```diff
@@ -28,14 +28,26 @@ on:
...
       distribution: otelcol-ebpf-profiler
       config_file: ebpf-profiler-config.yaml
-      docker_run_options: '--privileged --pid=host -v /sys/kernel/debug/:/sys/kernel/debug/:ro'
-      otelcol_run_options: '--feature-gates=service.profilesSupport'
+      docker_run_options: "--privileged --pid=host -v /sys/kernel/debug/:/sys/kernel/debug/:ro"
+      otelcol_run_options: "--feature-gates=service.profilesSupport"
       goos: '[ "linux" ]'
       goarch: '[ "amd64" ]'
     secrets: inherit
+  check-goreleaser-arm64:
```
douglascamata, line 42: The `arm64` build of the `ebpf-profiler` seem to only succeed on an `arm64` runner. I imagine this is related to kernel headers dependencies (most likely I guess) or something related to CGO (not sure).

.github/workflows/base-package-tests.yaml
mowies: this is a known issue, you can safely ignore it here :)

.github/workflows/base-package-tests.yaml
mowies, line 1: are the changes here even needed?

douglascamata: Good catch. I double checked some things here and saw that we only do package tests on the contrib distro and on linux/amd64, so I don't think these changes are required. I'll remove them.

.github/workflows/base-package-tests.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
 
   create-issue:
     name: Create GitHub Issue
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-22.04
```
mowies: this can stay at 24.04 since it's just for creating the GH issue
```suggestion
    runs-on: ubuntu-24.04
```

.github/workflows/ci-goreleaser-ebpf-profiler.yaml
mx-psi: @open-telemetry/ebpf-profiler-approvers could you help answer this?

florianl: For cross-compliation the environment variable `TARGET_ARCH=arm64` need to be set. Just setting `goarch` is not sufficient.

douglascamata: @florianl and would that work even in an amd64 host builder? It would be awesome if it did!

douglascamata: Asking because I see that the ebpf distro has CGO enabled and in this case you cannot simply cross-compile like a normal pure-Go project. It would require some hacks to cross-compile the C stuff.

douglascamata: @florianl another question: is the project even buildable without going through the Makefile? Seems like the ebpf profiler has a slightly complex build process that requires separate commands to build C libraries, right? I'm not sure if Go Releaser will be able to build this. I need to do some research.

florianl: > would that work even in an amd64 host builder? 

yes - it should work. With `make agent TARGET_ARCH=arm64` I cross-compile the project here on my amd64 host.

> ebpf distro has CGO enabled

I hope, I don't miss things, but the most recent tag of the project should not require CGO anymore.

> is the project even buildable without going through the Makefile?

Components, that are reachable via `go.opentelemetry.io/ebpf-profiler/collector` should be buildable without leveraging the Makefile.

>  Seems like the ebpf profiler has a slightly complex build process that requires separate commands to build C libraries, right? 

The C parts/eBPF parts do not need to be recompiled. Their artifacts are part of the git repository and embedded into the Go code via `//go:embed ebpf/tracer.ebpf.amd64`.

douglascamata: Hah! I think I got it right and the build's passing now. So feel free to disregard my questions unless you think there's something relevant there to be clarified.

mowies:
> I'm going to go with option 2, because I think the arm64 test is important, but I don't have a strong opinion here.

I agree

mowies (approved):
LGTM!
There's probably gonna be some follow ups from this anyways, so I'd say let's merge it and run a manual nightly release to check how things go

douglascamata:
> There's probably gonna be some follow ups from this anyways, so I'd say let's merge it and run a manual nightly release to check how things go

200%. I'm ready to handle any required quick follow up 💪

PR #1157, opened 2025-09-02
Switch from Prometheus self-scrape to OTLP telemetry

### LINKED ISSUE #836: Change the config.yaml of distributions to use the OTLP receiver instead of Prometheus
config.yaml contains a snippet to read internal metrics:
```
  # Collect own metrics
  prometheus:
    config:
      scrape_configs:
      - job_name: 'otel-collector'
        scrape_interval: 10s
        static_configs:
        - targets: ['0.0.0.0:8888']
```

It should be using OTLP now.

lakkawardhananjay:
- Removed Prometheus receiver configuration from:
  * distributions/otelcol/config.yaml
  * distributions/otelcol-contrib/config.yaml
  * tests/golden/config.yaml
- Added `service.telemetry.metrics` to expose internal metrics in OTLP
- Updated metrics pipeline to use `otlp` receiver and debug exporter

fixes #836


distributions/otelcol-contrib/config.yaml
```diff
@@ -47,6 +37,10 @@ exporters:
     verbosity: detailed
 
 service:
+  telemetry:
+    metrics:
+      level: basic
+      address: 0.0.0.0:8888
```
bogdandrutu: This does not work. See https://github.com/open-telemetry/opentelemetry-collector/blob/main/CHANGELOG.md#v1340v01280 and the `telemetry.disableAddressFieldForInternalTelemetry`

lakkawardhananjay: Thank you for the clarification @bogdandrutu .
With just `level: basic`, the metrics will be exposed on the default (localhost) endpoint, as outlined in the latest documentation. I'll recommit with the removed one accordingly.
Much appreciated!

tests/golden/config.yaml
```diff
@@ -1,12 +1,10 @@
+---
```
mx-psi:
```suggestion
```

Issue #1068, opened 2025-07-31
Remove the hard-coded version of go.opentelemetry.io/otel/exporters/prometheus when metric name suffix bug is fixed


songy23:
This is a placeholder to revert https://github.com/open-telemetry/opentelemetry-collector-releases/pull/1067 when the metric name suffix bug is fixed in go.opentelemetry.io/otel/exporters/prometheus

PR #1178, opened 2025-09-22
[distributions] Stop pinning Prometheus Go SDK dependency

mx-psi:
After open-telemetry/opentelemetry-collector-contrib/pull/42493 I think we are ready for this!

Fixes https://github.com/open-telemetry/opentelemetry-collector-releases/issues/1068

Fixes open-telemetry/opentelemetry-collector/issues/12939


songy23:
> https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/42475 already bumped the Prometheus dependency

That was caught by the replace in this repo so the artifacts are still on v0.58. Now v0.60 fixes it, we don't need to pin to v0.58 any more

songy23 (approved):

atoulme (approved):

PR #1195, opened 2025-09-30
chore: fix bump-versions

### LINKED ISSUE #1186: Release automation picks a wrong candidate version when version is left empty
E.g. https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/17947655417/job/51038001727

It tries to bump to 0.137.0 instead of 0.136.0 while current version is 0.135.0, with candidate versions leaving empty.

My guess is https://github.com/open-telemetry/opentelemetry-collector-releases/pull/1087 step 1 and 2 do not work well with each other. Step 1 already bumps version to N+1 and step 2 further bumps it to N+2

mowies:
This PR fixes an issue during release where the manifest version of distros was updated to one version to far.
Since the component versions are now bumped by renovate, the bump-versions script doesn't need to be concerned with them anymore and therefore needs to do less things in general.

Fixes #1186


songy23 (approved):
Thanks!

mowies:
cc @dmitryax in case something is weird during the release :)

PR #1198, opened 2025-10-03
Allow inplace upgrades of msi

### LINKED ISSUE #1197: Windows MSI does not allow upgrading in place
Hello!

I came across an issue while implementing https://github.com/voxpupuli/puppet-otelcol/pull/44 

Image

When otelcol msi is already installed, it is not possible to install the new version over it. I must admit that my Microsoft Knowledge isn't that great but from what I gathered the problem seems to be that the Id property set in https://github.com/open-telemetry/opentelemetry-collector-releases/blob/34b3c2f479541ba41561728a02ce11c91f7711b9/distributions/otelcol-contrib/windows-installer.wxs#L4 should actually be set to `Id="*"` to make the installer generate a new ID on each build.

When the UpgradeCode remains the same value, but the id changes,Windows should identify the package as an upgrade and it'll work.

cyberkov:
I hope this Fixes #1197 
Unfortunately I am not able to test this change locally (except for the make build, which doesn't seem to produce the msi).

### DIFF
--- a/distributions/otelcol-contrib/windows-installer.wxs
+++ b/distributions/otelcol-contrib/windows-installer.wxs
@@ -1,12 +1,12 @@
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
    <Product
       Name="OpenTelemetry Collector ({{ .Version }}) - {{ .Binary }} distribution"
-      Id="B250A214-D463-4E9B-8902-1DE5C19EA951"
+      Id="*"
       UpgradeCode="B7C263DD-95A5-436A-A025-DCA5200C2BE3"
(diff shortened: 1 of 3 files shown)


cyberkov:
Hello @pjanotti!

Thank you for the artifact. I successfully verified the in-place upgrade. The changelog file has been added to the PR.

pjanotti:
CI Failures in [CI - Core - GoReleaser / Package tests / Package Tests (rpm) (pull_request)](https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/18336421011/job/52221890462?pr=1198) are unrelated to the present changes.

PR #1235, opened 2025-10-31
[chore] Split up `cmd/goreleaser/internal/configure.go`

douglascamata:
Doing a little house keeping that I promised a very long time ago. This breaks up the huge `cmd/goreleaser/internal/configure.go` into many files, so that it's easy to find things and scroll around, particularly when it comes to looking at how we build the goreleaser configuration file for each one of the distros.

I'm very open to feedback on the names of the new files. Naming things is difficult. 

Small note: I generate the goreleaser files of all distros during my work to ensure that these changes still produces the exact same files as before.

PR #1260, opened 2025-11-13
[chore] Remove msi from opamp binary distribution

douglascamata:
On #1235 I crossed some wires and committed changes that were supposed to come in the future  to add an MSI installer to the `opampsupervisor` binary distribution (spoilers). 

The change included the Go code to generate a goreleaser config for the `opampsupervisor` to build the MSI installer (you can see it being removed in the diff). Luckily I never regenerated the goreleaser configs with this, so nothing broke and there is nothing to change there as well.

This unintended change passed CI because our `ensure-goreleaser-up-to-date` make target was only checking distributions and not binaries. I updated it to check binaries and prevent this issue from happening again.


Kingcb1:
Kingcb

PR #1285, opened 2025-12-02
[chore]: Upgrading go releases to latest major version

MovieStoreGuy:


MovieStoreGuy:
Sorry, I wanted to make sure both the upgrade and the required changes worked together.

> Is this the result of running make generate-goreleaser? 

So the changes are both the upgrade and running the expected command.

PR #1288, opened 2025-12-02
Fix: Trying to find issue in Ci

MovieStoreGuy:
# Context

I was looking to help resolve the issue of waiting for systemd to be running, however, as a work around it was easier to skip the check for now and look to address it later.

I've also included some additional logging on exit to capture any logs that are being produced by the container.


MovieStoreGuy:
Linking to issue: https://github.com/open-telemetry/opentelemetry-collector-releases/issues/1292

PR #1317, opened 2025-12-17
[chore] Remove replacement for openshift/api

halfdata:
Removes replacement for openshift/api as it's not needed any more. Version 3.9.0+incompatible was replaced by actual pseudo-version in https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/f3f32301c7a810c4cddf402cb04ecd9626a69d4d


distributions/otelcol-contrib/manifest.yaml
```diff
@@ -264,5 +264,3 @@ providers:
 
 # When adding a replace, add a comment before it to document why it's needed and when it can be removed
 replaces:
-  # see https://github.com/openshift/api/pull/1515
-  - github.com/openshift/api => github.com/openshift/api v0.0.0-20230726162818-81f778f3b3ec
```
dmathieu, line 268: This is being replaced in other distributions too (ebpf-profiler). Could we remove it there too?

distributions/otelcol-contrib/manifest.yaml
```diff
@@ -264,5 +264,3 @@ providers:
 
 # When adding a replace, add a comment before it to document why it's needed and when it can be removed
 replaces:
```
dmathieu: I don't think we need to keep this.

```suggestion
```

distributions/otelcol-contrib/manifest.yaml
halfdata: Done.

halfdata:
> please add a changelog entry for this, or add `[chore]` the the PR title to not show this in the changelog

Done.

songy23 (approved):

mowies (approved):

PR #1335, opened 2026-01-07
Prevent service being started/enabled on install for rpms

### LINKED ISSUE #1334: rpm package should not start or enable the service upon installation
The rpm package should not start or enable the service upon installation. There are a situations in which this will break, one of which I hit today, and it's generally unexpected for system administrators of rpm based distros.

In my specific example:
* Another service is usually configured to listen on one of the ports that otelcol-contrib listens on in the default config
* The host was rebuilt
* otelcol was re-installed, stealing the port from the other service
* Templating my otelcol config failed for unrelated reasons and thus that port was not freed up before the other service was started
* The other service failed to start

It's also against rpm packaging guidelines on [scriptlets](https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_systemd) and [default services](https://docs.fedoraproject.org/en-US/packaging-guidelines/DefaultServices/#_enabling_services_by_default), as the default configuration listens on several ports on 0.0.0.0.

(I note also that there are rpm features to handle the things these scripts do, but moving away from nFPM is a whole other kettle of fish, so I won't dwell on that here). 

I don't use any deb based distros, but I think starting packages is on install is more common there? Perhaps we need separate postinstall scripts for deb vs rpm? I am not familiar with nFPM but it does appear to support this from a quick read of [the docs](https://goreleaser.com/customization/nfpm/).

If you agree that this behaviour should change, I am willing to put up a PR.

sproberts92:
This brings behaviour inline with usual expectations for rpm packages which usually leave this up to the system administrator.

Create a separate postinstall-rpm.sh specifically for rpms and set this as an override for the postinstall script for rpm packages.

This resolves #1334.


.chloggen/rpm_dont_start_service.yaml
```diff
@@ -0,0 +1,26 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: "breaking"
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: "packaging"
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Prevent service being started/enabled upon install for rpms"
```
atoulme, line 10: this changelog doesn't reflect the changes. You seem to be checking for systemd presence in the system.
Please clarify this changelog.

sproberts92: I believe this is an accurate description of the change in behaviour for end users - please also refer to #1334 for more expansion on the motivation here.

All of the logic in the scripts that I am adding already exists in the base `postinstall.sh` scripts - my change merely creates specific versions with slightly different outcomes for rpms.

For example, for the plain otelcol distribution, the diff between the existing script and my rpm specific override:
```diff
--- a/distributions/otelcol/postinstall.sh
+++ b/distributions/otelcol/postinstall-rpm.sh
@@ -7,10 +7,9 @@ if command -v systemctl >/dev/null 2>&1; then
     if [ -d /run/systemd/system ]; then
         systemctl daemon-reload
     fi
-    systemctl enable otelcol.service
     if [ -f /etc/otelcol/config.yaml ]; then
         if [ -d /run/systemd/system ]; then
-            systemctl restart otelcol.service
+            systemctl try-restart otelcol.service
         fi
     fi
 fi
```

If you think there's a better approach then I'm happy to adjust.

distributions/otelcol-contrib/postinstall-rpm.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+if command -v systemctl >/dev/null 2>&1; then
+    if [ -d /run/systemd/system ]; then
+        systemctl daemon-reload
+    fi
+    if [ -f /etc/otelcol-contrib/config.yaml ]; then
```
atoulme: why are you checking for this file presence?

.chloggen/rpm_dont_start_service.yaml
atoulme: OK I see. There is boilerplate on that script you can strip away imo ; remove the check for the config file presence. Is checking for `/run/systemd/system` presence even necessary?

distributions/otelcol-contrib/postinstall-rpm.sh
sproberts92: Like the others, this was present in the original file that my rpm-specific version is derived from. However in this case as we're not attempting to start the service first time I agree that it's entirely useless, so I have removed it.

.chloggen/rpm_dont_start_service.yaml
sproberts92: I agree, have removed.

mowies (approved):

mowies:
i'm not an expert on RPM packages but I think this looks good overall

distributions/otelcol-contrib/postinstall-rpm.sh
mowies: @atoulme can this be resolved?

sproberts92: The line @atoulme was questioning has been removed so I think it's fair to resolve this thread.

sproberts92:
@mowies - what's the project etiquette on resolving the conversations? I left them open waiting for @atoulme to mark. But if he's not available any more to follow up, should we just merge it?

mowies:
sorry i was on PTO the last weeks. yes i think we can resolve the conversations and merge

PR #1343, opened 2026-01-13
[distro/ebpf-profiler] Add OTLP receiver

douglascamata:
This PR adds the OTLP Receiver to the `ebpf-profiler` distro. I would like to be able to use this receiver to monitor the Collector itself (internal telemetry), enrich it via the `resourcedetection` processor and re-export it via the OTLP exporter already present in the distro.

### DIFF
--- a/.chloggen/ebpf-profiler-otlp-receiver.yaml
+++ b/.chloggen/ebpf-profiler-otlp-receiver.yaml
@@ -5,7 +5,7 @@ change_type: enhancement
 component: ebpf-profiler
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Add the prometheus receiver to the ebpf-profiler distribution
+note: Add the otlp receiver to the ebpf-profiler distribution
 
 # One or more tracking issues or pull requests related to the change
 issues: [1343]
--- a/distributions/otelcol-ebpf-profiler/manifest.yaml
+++ b/distributions/otelcol-ebpf-profiler/manifest.yaml
@@ -27,7 +27,7 @@ processors:
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/processor/transformprocessor v0.143.0
 
 receivers:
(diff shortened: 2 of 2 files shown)


dmathieu:
Could you describe your use case a bit more?
You mention monitoring the collector itself. But you don't need prometheus to gather the collector's internal metrics. You can [export them directly with OTLP](https://opentelemetry.io/docs/collector/internal-telemetry/#otlp-exporter-for-internal-metrics).

douglascamata:
Hey @damemi, I'm interested in enriching the service telemetry with a well configured resource detection (cloud ones, k8s, custom resource attributes, etc), which is not available in the SDK. Even after https://github.com/open-telemetry/opentelemetry-collector/issues/14411 is solved, they might exist in some form, but won't be as rich in configuration options as the resource detection processor.

It also works for me if we add the otlp receiver to this distro, by the way. Not sure if this matter, just saying.

dmathieu:
How would the prometheus receiver (or OTLP for that matter) be needed to add additional resource detection?
You should be able to use any processor (that are included within the image) without any additional receivers.

dmathieu:
Sorry, not any processor, just the ones that support profiling (that will be mentioned in their README).
The [resourcedetectionprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor) does have profiles support. So adding it to your profiling receiver, like you would on an OTLP receiver should work.

Adding the OTLP receiver with processors will not help you get additional data onto your profiles, as the processors would run on entirely different pipelines, and the added data wouldn't be shared.

dmathieu:
Hmm so you want to have the collector send internal telemetry to itself using OTLP so it be enriched, and then exported elsewhere?

dmathieu:
I still don't think this is a good idea. The eBPF collector is meant to be a very lightweight DaemonSet collector that only handles profiles, and then exports the data to a cluser-wide collector.
With that architecture, the cluster-wide collector can receive your internal telemetry and augment its resources.

If we start doing this, we would open the door to adding a lot more things, such as exporters or other receivers, which defeats the purpose of the lightweight collector.

If you still wish to do this, you can always build your own distribution.

PR #1369, opened 2026-02-10
[chore] Add script to ensure version consistency

iblancasa:
This PR adds https://github.com/open-telemetry/opentelemetry-collector-releases/pull/1355 again but with some improvements to avoid causing issues due to the `renovate` PRs.

Tested locally.

PR #1378, opened 2026-02-18
[chore] Don't validate distro versions

evan-bradley:
Partially reverts #1369

I ran into an issue here: https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/22155025369/job/64056447144


evan-bradley:
The failure in "CI - Collector Binaries / Build (otelcol-contrib) (pull_request)" is expected, the component has been marked as unmaintained.

PR #1380, opened 2026-02-19
[chore] Force "GOOS=linux" when building otel-ebpf-profiler

ulyssessouza:
Very simple PR just to improve que experience of `make build` on macOS (instead of failing).
This just builds normally as a linux binary (still respects the current arch).

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -39,7 +39,13 @@ do
     echo "Using Builder: $(command -v "$BUILDER")"
     echo "Using Go: $(command -v go)"
 
-    if "$BUILDER" --skip-compilation="${skipcompilation}" --config manifest.yaml > _build/build.log 2>&1; then
+    build_env=""
+    if [[ "$distribution" = "otelcol-ebpf-profiler" ]]; then
+        build_env="GOOS=linux" # EBPF is linux exclusive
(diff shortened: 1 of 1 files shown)


douglascamata (changes requested):
@ulyssessouza could we also have the script output a small warning when it builds with a forced `GOOS=linux` so that people are not surprised afterwards with errors trying to running the binary? Otherwise this LGTM.

scripts/build.sh
```diff
@@ -35,11 +35,19 @@ do
...
+    build_env=""
+    if [[ "$distribution" = "otelcol-ebpf-profiler" ]]; then
+        build_env="GOOS=linux" # EBPF is linux exclusive
+    fi
+
     echo "Building: $distribution"
     echo "Using Builder: $(command -v "$BUILDER")"
     echo "Using Go: $(command -v go)"
+    if [[ ${#build_env} -gt 0 ]]; then
+        echo "⚠️ Using Environment Variables:" "${build_env[@]}"
```
ulyssessouza: @douglascamata It adds a warning mentioning that we are building with the (added) environment variable.
Please let me know if you would like to see anything extra

douglascamata: This warning is partially okay, because it tells what happened (a bit indirectly, but it does) but it doesn't say **why** and this is the most important part, imho.

PR #1395, opened 2026-03-01
feat: add /var/lib persistent state directories for collector RPM/DEB packages

### LINKED ISSUE #1396: Add /var/lib persistent state directories for collector RPM/DEB packages
## Summary

Add packaged `/var/lib/<package>` directories to `otelcol`, `otelcol-contrib`, and `otelcol-otlp` RPM/DEB distributions.

## Motivation

The [file_storage resiliency extension](https://opentelemetry.io/docs/collector/resiliency/) persists the collector queue to disk. When installed via RPM or DEB packages, the collector service user needs a writable directory under `/var/lib` to store this state. Without the directory being created and owned by the package, users must manually create it or configure an alternate path.

## Proposed Change

- Add `/var/lib/<package>` directory entry to the nfpm contents for each distribution
- Directory owned by the collector service user/group with mode `0750` (rwxr-x---)
- Affects: `otelcol`, `otelcol-contrib`, `otelcol-otlp`

dol:
RPM/DEB packages for `otelcol`, `otelcol-contrib`, and `otelcol-otlp` were missing `/var/lib/<package>` directories, preventing the [file_storage extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/filestorage) from persisting queue state on disk without manual post-install setup.
Extension is mentioned as part of [Resiliency](https://opentelemetry.io/docs/collector/resiliency/#persistent-storage-write-ahead-log---wal) improvements.

## Changes

- **`cmd/goreleaser/internal/builder.go`** — adds `withVarLibDir(user, group string)` helper that appends an nfpm `type: dir` content entry at `/var/lib/<package>`, owned by the service user/group at mode `0750`
- **`cmd/goreleaser/internal/distro_otelcol.go`**, **`distro_contrib.go`**, **`distro_otlp.go`** — each calls `withVarLibDir` with its respective service user/group
- **Generated `.goreleaser.yaml`** files for all three distributions updated accordingly

Resulting nfpm entry (example for `otelcol-contrib`):
```yaml
contents:
  - dst: /var/lib/otelcol-contrib
    type: dir
    file_info:
      owner: otelcol-contrib
      group: otelcol-contrib
      mode: 0750
```

Service users are created by the existing `preinstall.sh` scripts, so no additional scripting is required.


cmd/goreleaser/internal/builder.go
```diff
@@ -225,6 +225,23 @@ func (b *distributionBuilder) newNfpms(dist string) []config.NFPM {
...
+func (b *distributionBuilder) withVarLibDir(user, group string) *distributionBuilder {
+	b.configFuncs = append(b.configFuncs, func(d *distribution) {
+		for i := range d.Nfpms {
+			d.Nfpms[i].Contents = append(d.Nfpms[i].Contents, config.NFPMContent{
+				Destination: path.Join("/var", "lib", d.Name),
+				Type:        "dir",
+				FileInfo: config.FileInfo{
+					Owner: user,
+					Group: group,
+					Mode:  0750,
```
mowies: The mode looks correct here, but the generated goreleaser files have it as a decimal number which is not what we want, at least according to the [docs](https://goreleaser.com/customization/nfpm/?h=nfpm#:~:text=tree%0A%20%20%20%20%20%20%20%20file_info%3A%0A%20%20%20%20%20%20%20%20%20%20%23%20File%20mode.-,mode%3A%200644,-%23%20Modification%20time.%0A%20%20%20%20%20%20%20%20%20%20%23%0A%20%20%20%20%20%20%20%20%20%20%23%20Templates%3A%20allowed).
Please make sure this reaches the generated goreleaser files in the correct format.

dol: Agree. Thanks for the finding. I'll rework it.

dol: @mowies 
After looking into this more carefully, I'm not sure the effort is worth the cosmetic change here.

The `.goreleaser.yaml` files in this repo are **generated** by the Go tool in `cmd/goreleaser/`. The `Mode` field is stored as `os.FileMode` (`uint32`), and `go.yaml.in/yaml/v3` always serializes integers as decimal - there's no built-in option to emit octal notation. Making the generated output show `0o750` would require post-processing the encoder's stdout in `main.go` to do a string replacement, which feels like disproportionate complexity for a cosmetic concern.

`mode: 488` is **valid according to the [GoReleaser JSON schema](https://goreleaser.com/static/schema.json)** - the `mode` field is defined as `type: integer`, and `488` is exactly that. GoReleaser parses its config with `go.yaml.in/yaml/v3` (YAML 1.2), which reads `488` as decimal 488 - the same value as octal `0750`. The behavior is identical.

There is also a benefit around YAML version compatibility:

| Notation | YAML 1.1 (`yaml.v2`)  | YAML 1.2 (`yaml.v3`)           | Effective value (GoReleaser) |
| -------- | --------------------- | ------------------------------ | ---------------------------- |
| `0750`   | ✅ octal -> **488**   | ❌ decimal -> **750** (wrong!) | **750** (wrong!)             |
| `0o750`  | ❌ parse error        | ✅ octal -> **488**            | **488**                      |
| `488`    | ✅ decimal -> **488** | ✅ decimal -> **488**          | **488**                      |

`0o750` (YAML 1.2 octal syntax) would silently break under any YAML 1.1 parser, and `0750` (YAML 1.1 octal syntax) parses as decimal `750` under YAML 1.2 - wrong and silent. Plain decimal `488` is unambiguous in **both** YAML versions.

Is keeping `mode: 488` acceptable given it is schema-valid, functionally correct, and unambiguous across YAML versions?

mowies: ok i guess then leaving it in the decimal format is also fine. let's maybe add a comment in the code though to help confused readers 😅

dol: Comment was added.

mowies (approved):

mowies:
Windows builds are still failing, but this has nothing to do with Windows whatsoever so I think it's fine to merge

PR #1405, opened 2026-03-06
[chore] add explicit telemetry configuration to config.yaml files

atoulme:
Expose explicitly the default telemetry settings in configuration.

This is a no-op change, since this matches the default configuration. It reduces user confusion as to where we bind to port 8888.

### DIFF
--- a/distributions/otelcol/config.yaml
+++ b/distributions/otelcol/config.yaml
@@ -55,6 +55,11 @@ service:
               prometheus:
                 host: '127.0.0.1'
                 port: 8888
+                without_scope_info: true
+                without_type_suffix: true
+                without_units: true
+                with_resource_constant_labels:
+                  included: [ ]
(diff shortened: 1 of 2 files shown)


distributions/otelcol/config.yaml
```diff
@@ -47,6 +47,14 @@ exporters:
...
 
 service:
+  telemetry:
+    metrics:
+      readers:
+        - pull:
+            exporter:
+              prometheus:
+                host: '127.0.0.1'
+                port: 8888
```
bogdandrutu, line 57: Please use also:

```
                    without_scope_info: true
                    without_type_suffix: true
                    without_units: true
                    with_resource_constant_labels:
                      included: []
```

atoulme: added

PR #1406, opened 2026-03-09
[chore] Delete old nightly tags and images

### LINKED ISSUE #1088: [Nightlies] Clean up older tags and images
To not keep too many nightly versions into the past, both tags and images could be cleaned up.

Following what is done in [this Goreleaser blog post](https://goreleaser.com/blog/nightly-actions/#dont-forget-to-clean-up) should work here as well.

mowies:
#### Description

Adds a `cleanup-nightly` reusable workflow that's also added to the nightly release workflow. After each successful nightly build, the job deletes git tags, GHCR package versions, and DockerHub tags older than 2 weeks.
The workflow can also be triggered manually and there is a dry-run option.

  #### Link to tracking issue

  Fixes #1088

  #### Authorship

#### AI Disclosure
This PR was assisted by Claude.


.github/workflows/nightly-release.yaml
```diff
@@ -41,3 +41,152 @@ jobs:
...
+          permission-packages: write
+
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+          token: ${{ steps.otelbot-token.outputs.token }}
+
+      - name: Delete nightly git tags older than 2 weeks
+        run: |
+          CUTOFF=$(date -d '2 weeks ago' +%s)
```
dmitryax: Why '2 weeks ago' specifically? Seems to be pretty aggressive given that we haven't removed any of them before.

Also, I guess the initial run would likely to time out 😅

mowies: No specific reason honestly other than having a somewhat clean list of tags in the repo. This just seemed like a reasonable TTL to me since we create tags every night.
The first few runs would probably run into issues but I hope that will clear up after a few runs when the deletion step has caught up. But I can look into some deletion limit if you want, maybe 2 weeks again 😄

mowies: I will limit the deletion timeframe to 2 weeks worth of tags and images, plus the workflow will get a manual trigger as well to clean up more tags manually at the start.

mowies: I will also add an option for a dry-run that just prints out the list of tags to delete

mowies: ok i think i'm done with all the changes now 👍🏼 it's now a reusable workflow that can also be triggered manually to expedite cleanup if we want to and to also allow for manual dry runs.
i left the 2 weeks limit as is, i think it should be fine and will also prevent a huge run when this is merged.

douglascamata (approved):

PR #1414, opened 2026-03-13
[wip] Add windows nightlies

### LINKED ISSUE #1091: [Nightlies] Add windows nightly releases
Windows is currently excluded from nightly releases since there were some issues during initial implementation.

Look into those issues and add Windows to the nightly releases.

douglascamata:
This PR attempts to add Windows nightlies.

It was previously attempted but there were some failures because of strict versioning requirements on the `wxs` files. 

Here we try to set the nightly version to: `{{ .RawVersion }}.{{ .Now.Format "00215" }}`. 

According to https://pkg.go.dev/time, the time format `00215` will give us: zero-padded day of the year and hour in 24-hour format. This should be enough for the nightlies, as they run only once per day. 

This should close #1091.

### DIFF
--- a/.github/workflows/release-contrib.yaml
+++ b/.github/workflows/release-contrib.yaml
@@ -20,12 +20,12 @@ jobs:
     permissions: write-all
   release-windows:
     name: Release Contrib (Windows)
-    if: ${{ !contains(github.ref, '-nightly') }}
     uses: ./.github/workflows/base-release.yaml
     with:
       distribution: otelcol-contrib
       goos: '[ "windows" ]'
       goarch: '[ "386", "amd64", "arm64" ]'
       runner_os: windows-2022
+      nightly: ${{ contains(github.ref, '-nightly') }}
     secrets: inherit
     permissions: write-all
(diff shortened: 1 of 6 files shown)


.github/workflows/release-contrib.yaml
```diff
@@ -20,12 +20,12 @@ jobs:
...
   release-windows:
     name: Release Contrib (Windows)
-    if: ${{ !contains(github.ref, '-nightly') }}
     uses: ./.github/workflows/base-release.yaml
     with:
       distribution: otelcol-contrib
       goos: '[ "windows" ]'
       goarch: '[ "386", "amd64", "arm64" ]'
       runner_os: windows-2022
+      nightly: ${{ contains(github.ref, '-nightly') }}
```
mowies, line 29: was this change just for testing?

douglascamata: Nope. According to add what I see above on we will also need this here, no? I didn't have the chance to run some tests from my own fork yet, but I will.

mowies:
@douglascamata is this relevant? was it done already?

douglascamata:
Oh wow, I didn't notice this was about to get auto closed, thanks for the ping @mowies. I think it's still relevant to help show that Windows is strongly supported as tier 2 platform. I'm a bit short on time to test the nightlies for now, but I will try to get back to it.

PR #1424, opened 2026-03-18
[all] add aix/ppc64 binaries

atoulme:
See https://github.com/open-telemetry/opentelemetry-collector/issues/13380 for context.


mowies:
@atoulme this seems to still be relevant, but also the blocking PR is in draft mode and another one got closed for inactivity. Shall we keep this PR open for now or is it fine to close?

atoulme:
This is no longer blocked. Reopening.

songy23 (commented):
looks like you need to exclude components that doesn't support aix

atoulme:
> looks like you need to exclude components that doesn't support aix

that should not be needed, but the build might not be catching the latest contrib.

distributions/otelcol-contrib/aix.dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM --platform=linux/amd64 alpine:3.23@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11 as certs
```
mowies: couldn't this be solved using goreleaser directly without needing separate dockerfiles?
goreleaser has docs on it: https://goreleaser.com/resources/cookbooks/multi-platform-docker-images/#creating-multi-platform-docker-images-with-goreleaser

atoulme: good find, I can use to reduce the diff. That said, not sure this works so well if we have a multi-stage builder.

atoulme: actually, see, we set it here: https://github.com/open-telemetry/opentelemetry-collector-releases/pull/1424/changes#diff-67f62a17add1bfe19f5be6dcb9010b87b6a98408431b6dbf10899da1bce72b05R106

The problem is in the multi-stage build, the first step must be done under linux/amd64 because alpine has no aix build, and then use the default `--platform` tag when building from scratch.

atoulme:
@Dylan-M please review

.github/workflows/base-ci-goreleaser.yaml
```diff
@@ -71,8 +87,14 @@ jobs:
             GOARCH: arm
           - GOOS: windows
             GOARCH: arm
+          - GOOS: windows
+            GOARCH: aix
```
Dylan-M: aix isn't a valid entry for GOARCH.

Based on entry position, I'm guessing this may have been intended as either ppc64le and/or riscv64?

.github/workflows/base-ci-goreleaser.yaml
```diff
@@ -278,6 +300,15 @@ jobs:
           load: true
           tags: ${{ env.TAG }}
 
+      - name: Build container images locally (AIX)
```
Dylan-M: AIX is not valid in docker.  Does this step build AIX WPARs? I didn't think that was a thing in this toolchain.

.github/workflows/release-k8s.yaml
```diff
@@ -13,8 +13,8 @@ jobs:
     uses: ./.github/workflows/base-release.yaml
     with:
       distribution: otelcol-k8s
-      goos: '[ "linux" ]'
-      goarch: '[ "amd64", "arm64", "ppc64le", "riscv64", "s390x" ]'
+      goos: '[ "aix", "linux" ]'
```
Dylan-M: Same concern about AIX + Docker Images as previously mentioned.

distributions/otelcol-k8s/.goreleaser.yaml
```diff
@@ -11,13 +11,25 @@ env:
 release:
   replace_existing_artifacts: true
 builds:
+  - id: otelcol-k8s-aix
```
Dylan-M: Same docker concern.

distributions/otelcol-k8s/aix.dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM --platform=linux/amd64 alpine:3.23@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11 as certs
```
Dylan-M: This looks like it is building a Linux docker container, and I'll echo all of my previous docker + aix concerns.

distributions/otelcol-otlp/aix.dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM --platform=linux/amd64 alpine:3.23@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11 as certs
```
Dylan-M: Docker/AIX again

distributions/otelcol/.goreleaser.yaml
```diff
@@ -140,6 +152,26 @@ nightly:
   tag_name: nightly-otelcol
   keep_single_release: true
 dockers:
+  - goos: aix
+    goarch: ppc64
+    dockerfile: aix.dockerfile
```
Dylan-M: AIX + Docker again

distributions/otelcol/aix.dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM --platform=linux/amd64 alpine:3.23@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11 as certs
```
Dylan-M: One more AIX + Docker

Dylan-M (changes requested):
AIX is not valid for Docker, and your dockerfiles for it all look like they build linux/amd64 images anyway.

.github/workflows/base-ci-goreleaser.yaml
atoulme: this is an exclusion. We exclude a number of combinations from the job.

.github/workflows/base-ci-goreleaser.yaml
atoulme: OK I will remove the docker image build.

distributions/otelcol-k8s/aix.dockerfile
atoulme: this is a multistage build where the first image is using linux/amd64, and the final image reuses a file from it.

.github/workflows/ci-goreleaser-k8s.yaml
```diff
@@ -36,8 +36,8 @@ jobs:
     uses: ./.github/workflows/base-ci-goreleaser.yaml
     with:
       distribution: otelcol-k8s
-      goos: '[ "linux" ]'
-      goarch: '[ "amd64", "arm64", "ppc64le", "riscv64", "s390x" ]'
+      goos: '[ "aix", "linux" ]'
+      goarch: '[ "amd64", "arm64", "ppc64", "ppc64le", "riscv64", "s390x" ]'
```
mowies: does the k8s distro make sense if there's no container builds?

.github/workflows/release-k8s.yaml
```diff
@@ -13,8 +13,8 @@ jobs:
...
+      goarch: '[ "amd64", "arm64", "ppc64", "ppc64le", "riscv64", "s390x" ]'
```
mowies: same concern here

.github/workflows/base-ci-goreleaser.yaml
```diff
@@ -63,6 +63,22 @@ jobs:
         GOOS: ${{ fromJSON( inputs.goos) }}
         GOARCH: ${{ fromJSON( inputs.goarch) }}
         exclude:
+          - GOOS: aix
```
mowies: at this point i think it would be easier to use an `include` instead of 10 `excludes` 😅

.github/workflows/base-ci-goreleaser.yaml
```diff
@@ -310,11 +332,52 @@ jobs:
...
```
mowies: same here with includes vs excludes

.github/workflows/ci-goreleaser-k8s.yaml
atoulme: Right - I'll remove.

.github/workflows/base-ci-goreleaser.yaml
atoulme: My github actions foo is a bit rusty, is that really what you want, or is there a way to make this more elegant?

.github/workflows/base-ci-goreleaser.yaml
Dylan-M: > this is an exclusion. We exclude a number of combinations from the job.

Right, but AIX is a GOOS, not a GOARCH, so this is still an invalid combination.

.github/workflows/base-ci-goreleaser.yaml
mowies: i think it would be enough to add 
```
include:
  - GOOS: aix
    GOARCH: ppc64
```

and remove all the excludes

.github/workflows/ci-goreleaser-contrib.yaml
```diff
@@ -40,6 +40,15 @@ jobs:
...
     secrets: inherit
 
+  check-goreleaser-aix:
+    name: CI - Contrib - GoReleaser
+    uses: ./.github/workflows/base-ci-goreleaser.yaml
+    with:
+      distribution: otelcol-contrib
+      goos: '[ "aix" ]'
+      goarch: '[ "ppc64" ]'
+    secrets: inherit
```
douglascamata, line 50: Any good reason to not fold the aix/ppc64 target from here into the job above? It already includes checks for Linux and Darwin OSes, so adding the AIX should be fine. The Windows one is only separate because it requires a different runner OS.

This applies in all the distros CI and also to the distro release workflows (i.e. `release-contrib.yaml` and so on).

douglascamata (approved):
I have one non-blocking and non-critical question, but LGTM.

.github/workflows/release-contrib.yaml
```diff
@@ -18,6 +18,17 @@ jobs:
...
     secrets: inherit
     permissions: write-all
+  release-aix:
+    name: Release Contrib (AIX)
+    if: ${{ !contains(github.ref, '-nightly') }}
+    uses: ./.github/workflows/base-release.yaml
+    with:
+      distribution: otelcol-contrib
+      goos: '[ "aix" ]'
+      goarch: '[ "ppc64 ]'
```
douglascamata: Small typo here. 

```suggestion
      goarch: '[ "ppc64" ]'
```

douglascamata (changes requested):
Found some important things on a second look:

- In `.github/workflows/base-release.yaml` (untouched in this PR), you need to add a step to download the artifacts for AIX before the step named `Display structure of downloaded files` (L255). The artifacts are uploaded by another job and downloaded there using a name pattern that includes the OS, so it ignores the AIX artifacts. 

- In `.github/workflows/release-core.yaml` and `.github/workflows/release-otlp.yaml` (also untouched in this PR) all you need to add the `aix/ppc64` release trigger for these distros. Basically the same thing you did here in `.github/workflows/release-contrib.yaml`.

- There's a typo that will cause a build failure on release.

In general, I would recommend to try running these release workflows in a fork. It requires some changes to container image repo references and a GoReleaser Pro key.

.github/workflows/ci-goreleaser-contrib.yaml
mowies: agreed

distributions/otelcol-contrib/.goreleaser.yaml
```diff
@@ -18,6 +18,20 @@ msi:
...
+  - id: otelcol-contrib-aix
+    goos:
+      - aix
+    goarch:
+      - ppc64
+    goarm:
+      - "7"
+    goppc64:
+      - power8
+    dir: _build
```
Dylan-M, line 30: nit: `goarm` and `goppc64` don't apply to an AIX/ppc64 build. The AIX entries in `distributions/otelcol/.goreleaser.yaml` and `distributions/otelcol-otlp/.goreleaser.yaml` (and the build-config in this distro's own `.goreleaser-build.yaml`) don't set these. Looks like copy-paste leftover from the linux block — safe to drop.

.chloggen/add_aix.yaml
```diff
@@ -0,0 +1,26 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
Dylan-M, line 25: Question: was `change_logs: []` intentional? Empty excludes this entry from every changelog. For new user-visible binaries `[user]` is what the template default suggests — wondering if you wanted that here.

Dylan-M (approved):

.chloggen/add_aix.yaml
atoulme: no, if empty, it will go to the user changelog. You can see it here: https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/.chloggen/config.yaml#L23

There is only one type of changelog in this repository.

.github/workflows/release-contrib.yaml
atoulme: good catch, fixing

distributions/otelcol-contrib/.goreleaser.yaml
atoulme: those files are generated. The same keys are present for all build types, including Windows distributions.

.github/workflows/ci-goreleaser-contrib.yaml
atoulme: that's to avoid having a big exclude block as in https://github.com/open-telemetry/opentelemetry-collector-releases/pull/1424#discussion_r3279239062

atoulme:
I have made the changes recommended. I don't have access to a GoReleaser Pro key, unfortunately.

.github/workflows/release-k8s.yaml
douglascamata, line 1: I think this was changed by accident. You are adding aix/ppc64 to core, contrib and otlp. This file is for the k8s distro. The `release-core.yaml` workflow is still missing changes for this new target.

douglascamata (changes requested):

douglascamata:
> I have made the changes recommended. I don't have access to a GoReleaser Pro key, unfortunately.

@atoulme: If you can't test the `aix/ppc64` release, I think it would be a good idea to enable nightlies for it. Without this we will only know if all these changes work next time a release is due. If you just remove `if: ${{ !contains(github.ref, '-nightly') }}` from the release workflow for this target, it will have nightlies and we will find out any potential issues before the next scheduled release.

atoulme:
> > I have made the changes recommended. I don't have access to a GoReleaser Pro key, unfortunately.
> 
> @atoulme: If you can't test the `aix/ppc64` release, I think it would be a good idea to enable nightlies for it. Without this we will only know if all these changes work next time a release is due. If you just remove `if: ${{ !contains(github.ref, '-nightly') }}` from the release workflow for this target, it will have nightlies and we will find out any potential issues before the next scheduled release.

Briliant, thank you! Will try that.

douglascamata:
Looks like `github.com/cockroachdb/pebble/v2/vfs` isn't building on `aix/ppc64`:

> Error:     │ /home/runner/go/pkg/mod/github.com/cockroachdb/pebble/v2@v2.1.6/vfs/vfs.go:235:18: cannot use defaultFS{} (value of struct type defaultFS) as FS value in variable declaration: defaultFS does not implement FS (missing method GetDiskUsage)

luke6Lh43:
Thanks everyone for working on this PR - looking forward to seeing AIX support land!

atoulme:
> Looks like `github.com/cockroachdb/pebble/v2/vfs` isn't building on `aix/ppc64`:
> 
> > Error:     │ /home/runner/go/pkg/mod/github.com/cockroachdb/pebble/v2@v2.1.6/vfs/vfs.go:235:18: cannot use defaultFS{} (value of struct type defaultFS) as FS value in variable declaration: defaultFS does not implement FS (missing method GetDiskUsage)

Looks like that should be caught in the contrib repository. I'll go check there.

atoulme:
@douglascamata please take another look when you get a chance, I'd love to get this in before the next release

distributions/otelcol-contrib/.goreleaser.yaml
douglascamata: @atoulme I think we might need to do a change here in the checksum filename template. 

The macOS/linux build will end up producing `opentelemetry-collector-releases_otelcol-contrib_checksums.txt` and the `iax` build will produce the same file name, but in a different runner. So when creating the release artifacts one will overwrite the other. I suggest that we change this checksum filename template to this:

```yaml
checksum:
  name_template: '{{ .ProjectName }}_otelcol-contrib{{ if in (list "windows" "aix") .Runtime.Goos }}_{{ .Runtime.Goos }}{{ end }}_checksums.txt'
```

Small note: I just discovered that `in` template function to check if a slice contains a value and mostly copied the official example: https://goreleaser.com/customization/general/templates/#functions-pro.

douglascamata (changes requested):
@atoulme I took another look at this and besides this very last comment it's looking good!

distributions/otelcol-contrib/.goreleaser.yaml
atoulme: Good catch thanks! I have edited the template.

douglascamata (approved):

PR #1470, opened 2026-04-21
[chore] fix broken pipe errors in CI pipeline

mowies:
**Fix intermittent CI failures caused by broken pipe in `validate-components.sh`**

`printf '%s\n' "$valid_components" | grep -qxF "$component"` was intermittently failing under `set -o pipefail`. When `grep -q` finds a match it exits early, closing the pipe while `printf` is still writing — causing `printf` to receive `SIGPIPE` (exit 141). With `pipefail`, the pipeline returns 141 (non-zero), and the surrounding `if !` negates it to _true_, incorrectly treating valid components as missing and failing the build.

Fixed by replacing the pipe with a here-string (`<<< "$valid_components"`), which passes the data directly to grep with no pipe involved.

**Example failing runs:**
- https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/24708467235
- https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/24699008332
- https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/24698836287

This was implemented using Claude Code.

### DIFF
--- a/.github/workflows/base-ci-goreleaser.yaml
+++ b/.github/workflows/base-ci-goreleaser.yaml
@@ -98,6 +98,11 @@ jobs:
         with:
           platforms: arm64,ppc64le,linux/arm/v7,s390x,riscv64
 
+      - name: Start Docker daemon (Windows)
+        if: runner.os == 'Windows'
+        shell: powershell
+        run: Start-Service docker
+
       - name: Setup wixl # Required to build MSI packages for Windows
         if: runner.os != 'Windows'
         run: |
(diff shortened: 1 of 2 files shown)


mowies:
the fix from the last commit seems to work here as well 👍🏼 
reverting, since it's content from a different PR

woodpeqr (approved):

douglascamata (approved):
This intermittent failure has been bugs us since a long time. I'm more than happy if this solves the issue.

PR #1471, opened 2026-04-21
[chore] explicitly start Docker daemon on Windows runners

mowies:
Some Windows-2022 GitHub Actions VM instances don't have the Docker daemon running by default, causing intermittent failures in both check-goreleaser (GoReleaser docker build) and docker-tests (docker image load). Both jobs now explicitly call Start-Service docker before any Docker operations.

This was implemented using Claude Code.


mowies:
this fix also worked in https://github.com/open-telemetry/opentelemetry-collector-releases/pull/1470

dmathieu (approved):

PR #1553, opened 2026-07-10
[contrib] add the eBPF profiler receiver to the contrib distribution

### LINKED ISSUE #1552: Consider adding eBPF profiler receiver to the contrib distribution
I would like to see the eBPF profiler receiver to the contrib distribution for multiple reasons:
* It would get more adoption
* It would allow to check if the receiver works well on unsupported OSes (as in, doesn't crash compilation)
* Help external vendors, selfishly, adopt it as well in their own distribution knowing this is tested via contrib.

atoulme:
#### Description
Add the EBPF profiler receiver to the contrib distribution

#### Link to tracking issue

Fixes #1552 

#### Authorship

### DIFF
--- a/.chloggen/add_ebpf_profiler.yaml
+++ b/.chloggen/add_ebpf_profiler.yaml
@@ -1,0 +1,26 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: contrib
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add eBPF profiler receiver to the contrib distribution
+
+# One or more tracking issues or pull requests related to the change
+issues: [1552]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
+
--- a/distributions/otelcol-contrib/manifest.yaml
+++ b/distributions/otelcol-contrib/manifest.yaml
@@ -136,6 +136,8 @@ processors:
 receivers:
   - gomod: go.opentelemetry.io/collector/receiver/nopreceiver v0.156.0
   - gomod: go.opentelemetry.io/collector/receiver/otlpreceiver v0.156.0
+  - gomod: go.opentelemetry.io/ebpf-profiler v0.0.202627
+    import: go.opentelemetry.io/ebpf-profiler/collector
   # OBI (eBPF Instrumentation)
   - gomod: go.opentelemetry.io/obi v0.10.0
     import: go.opentelemetry.io/obi/collector


MrAlias:
Looks like the profiler requires `github.com/cilium/ebpf v0.22.0`, while OBI v0.10.0 expects v0.20.0. Go selects v0.22.0, causing OBI to fail on the removed `btf.FlushKernelSpec` symbol. OBI on `main` now supports `cilium/ebpf v0.22.0`, so this conflict should be resolved once we consume an OBI release containing that update.

MrAlias:
Also, the profiler does not currently compile on Windows; it references Linux-only `syscall.Mmap` APIs. Since `otelcol-contrib` targets Windows, the component will need appropriate platform guards or another way to exclude it from unsupported targets.

cc @open-telemetry/ebpf-profiler-approvers

douglascamata:
> Aslo, the profiler does not currently compile on Windows; it references Linux-only `syscall.Mmap` APIs. Since `otelcol-contrib` targets Windows, the component will need appropriate platform guards or another way to exclude it from unsupported targets.
> 
> cc @open-telemetry/ebpf-profiler-approvers

So this PR remains blocked until this problem is fixed, otherwise the Windows binary will start breaking at runtime. Could we have an issue opened in the correct repository and link it here, please?

MrAlias:
> Could we have an issue opened in the correct repository and link it here, please?

Fix: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1632

douglascamata:
@MrAlias thanks! So that should also allow the component to start in Windows without crashing, regardless of wether eBPF profiling works in Windows, right? I don't know how's the state of Windows support for this, I just want to ensure it won't crash at runtime. 😄

MrAlias:
> @MrAlias thanks! So that should also allow the component to start in Windows without crashing, regardless of wether eBPF profiling works in Windows, right? I don't know how's the state of Windows support for this, I just want to ensure it won't crash at runtime. 😄

From my review and testing, it won’t crash. If configured on Windows, receiver creation returns a clear unsupported-platform error and the Collector exits cleanly, similar to OBI.

@florianl can you confirm this behavior?

douglascamata:
@MrAlias if it behaves as you say, it sounds good to me. Waiting for confirmation.

Gandem:
As discussed in https://cloud-native.slack.com/archives/C03J794L0BV/p1783965656082109, I’m requesting that we hold this PR until we have documented and reviewed its security model.

This is not just a question of whether the receiver compiles. In daemonset mode, the profiler requires host process visibility and elevated capabilities for eBPF and process inspection (including reading the process' memory, and access to the entire filesystem of the host). These privileges are granted to the entire collector process, not isolated to the profiler receiver. If the same process also runs network-facing receivers, extensions, or other contrib components, an RCE in any one of them could inherit those ambient permissions and potentially lead to host-level compromise.

Packaging the receiver in contrib does not itself grant those privileges, but it makes running a broad, “kitchen sink” distribution with them an expected deployment path. Before we enable that path, I think we need:
1. A documented threat model covering the attack surface, and the impact of any component being compromised.
2. A documented least-privilege deployment, with the minimum required permissions for each supported mode.
3. Agreement from the Profiling SIG and the relevant Collector/release maintainers, with input from OBI and the TC/security stakeholders, on the security posture we want to support.

To be clear, I am not arguing that the profiler receiver can never be included in contrib (I definitely support portability and would love to see a path to contrib inclusion), nor am I asking that the cross-compilation work stop. 

I’m asking that we resolve and document these distribution-level security implications before merging the inclusion change.

I will be raising this during the next Profiling SIG meeting to get broader input on this and discuss next steps.

florianl:
> From my review and testing, it won’t crash. If configured on Windows, receiver creation returns a clear unsupported-platform error and the Collector exits cleanly, similar to OBI.
> 
> @florianl can you confirm this behavior?

Yes, this is the expected behavior and I can confirm the same in my tests here.

eBPF-based observability, like profiling and OBI, fundamentally change the privilege model of the collector: both require elevated capabilities and broad host visibility that go well beyond what a typical otelcol-contrib deployment needs. Bundling them into the same process as general-purpose contrib components means those ambient privileges are inherited by every other receiver, exporter, and extension running alongside, which significantly widens the blast radius of any vulnerability.

I think, this raises a broader architectural question worth discussing: should there be distinct, purpose-built distributions? One optimized for daemonset deployments (where OBI and profiling receivers can run with the necessary privileges in an appropriately scoped process) and one for sidecar/general-purpose use (where those privileges are neither needed nor desirable)? A "kitchen sink" contrib distribution that implicitly supports both deployment models without clear guidance seems to be the wrong default to me.

Like @Gandem I'm not arguing against profiler inclusion in contrib long-term, but I do think the threat model and least-privilege deployment guidance @Gandem is asking for are prerequisites, not just for security reasons, but so that users understand what they're opting into when they enable this receiver.

douglascamata:
I don't think we need purpose built distributions. In the end, security is controlled from the tools handling the deployment model like Helm charts, deb/rpm packages and the systemd service configuration, etc. Here we only build binaries and packages.

For example, it's in a Helm chart that you could configure how you run your Collector. You could run things like this in k8s (could be adapted for other platforms) to keep permissions tight and reduce the attack surface:

- One daemonset of the Collector Contrib for general things (i.e. everything except ebpf-profiler)
- Another one with high privileges that runs just the eBPF profiler receiver and has an OTLP exporter. Then it exports data to the "general" one, which in turn processes and exports that data to whatever backend you want.

So besides working on presets in the official Helm chart, maybe there should be some notes in official docs about recommended deploy models?

mowies:
I would argue that use-case specific distributions could make a lot of sense for certain cases and we do have some already (k8s distro for example).

What I see as an issue here is that otelcol-contrib is the de-facto go-to Collector that is recommended and that a new user would use as a start. I don't think it is wise to change the attack surface of such a widely-used component after the fact.


> * Another one with high privileges that runs just the eBPF profiler receiver and has an OTLP exporter. Then it exports data to the "general" one, which in turn processes and exports that data to whatever backend you want.

As @douglascamata already mentions, a production deployment of this would probably be a purpose-built Collector binary for just this use case to limit the blast radius. I'd say if this is already the first solution that comes to mind, it would make sense to also think about recommending that from the OTel Collector side.

Gandem:
I agree that the two-Collector deployment described above would be a good setup 👍 However, the Helm chart (to take one example), does not currently create or enforce that model.

The [image check](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/8b77b9a7b7e1b20f51cb507bece44328ba041960/charts/opentelemetry-collector/templates/NOTES.txt#L5-L18) we currently have only verifies that the selected image contains the profiling receiver (so currently the curated eBPF profiling distribution). Enabling the preset then:

- [merges profiling into the existing base configuration](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/8b77b9a7b7e1b20f51cb507bece44328ba041960/charts/opentelemetry-collector/templates/_config.tpl#L69-L99), without removing other pipelines or preventing other presets;
- leaves the [default network-facing OTLP, Jaeger, Zipkin, and Prometheus receivers](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/8b77b9a7b7e1b20f51cb507bece44328ba041960/charts/opentelemetry-collector/values.yaml#L208-L311) enabled;
- [runs the entire container as root and privileged by default](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/8b77b9a7b7e1b20f51cb507bece44328ba041960/charts/opentelemetry-collector/templates/_pod.tpl#L30-L39), with [host PID enabled](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/8b77b9a7b7e1b20f51cb507bece44328ba041960/charts/opentelemetry-collector/templates/daemonset.yaml#L40-L46).

Users can still manually create a dedicated profiling Collector, but with default values, the preset produces a privileged, general-purpose Collector. This is slightly less concerning with the dedicated image because it ships a smaller set of components than contrib.

Before allowing contrib through the image check, I think we should have an official, tested minimal configuration or a fail-closed isolation mode that makes the recommended deployment clear and **prevents accidental privilege sharing.**

> I don't think we need purpose built distributions. In the end, security is controlled from the tools handling the deployment model like Helm charts, deb/rpm packages and the systemd service configuration, etc. Here we only build binaries and packages.

I think this is a broader design question we should figure out, i.e. should this boundary be enforced:
* only at the deployment/configuration level, by running a narrowly configured contrib instance? 
* or also at the binary/distribution level, by limiting the components shipped in the privileged process?

Enforcing this only at the deployment/configuration level gives users more flexibility. A dedicated distribution, though, means fewer components are included / fewer code is running in the privileged binary. I think we should make that trade-off explicitly as part of the threat-model discussion.

That’s why I’d like the threat model and least-privilege deployment guidance to be in place before adding components that require privileged access to contrib.

douglascamata:
> A narrowly configured contrib deployment gives users more flexibility, while a purpose-built distribution reduces the amount of code running in the privileged process. I think we should make that trade-off explicitly as part of the threat-model discussion.

If users want a narrow distribution for k8s, we have the k8s distribution.

If users want a narrow distribution for data export, we have the otlp distro.

If users want a narrow, generic distribution, there's core.

I don't think contrib's intention is to be narrow.

Buuuut, if we do want contrib to be narrow, this component can be added in the `ebpf-profiler` distro (if it's not already there).

douglascamata:
Expanding a bit on the purpose of the Contrib distro, I pulled this from [its own README](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-contrib#recommendation):

> 
> ## Recommendation
> 
> As this distribution contains many components, it is a good starting point to try various configurations. However, when running in production, it is recommended to limit the collector to contain only the components necessary for an environment. Some reasons to do this:
> 
> * reduce the size of the collector, reducing deployment times for the collector
> * improve the security of the collector by reducing the available attack surface area
> 
> Building a [custom collector](https://opentelemetry.io/docs/collector/custom-collector/) can be achieved using the [OpenTelemetry Collector Builder](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder).

Gandem:
That’s fair, and I agree that contrib isn’t intended to be narrow. But I don’t think “wide” necessarily means there should be no boundaries. 

There is a difference between including many components and including components whose expected daemonset deployment requires host PID visibility and powerful Linux capabilities. Components requiring this level of privilege are relatively new in contrib, and they materially change the impact of a vulnerability in anything running alongside them.

The recommendation to build a custom distribution for production is useful, but it doesn’t match what I usually see in practice. Most commercial deployments I’ve encountered use one of the published distributions (or a vendor distribution) rather than maintaining their own Collector build. That makes the defaults and guardrails in those distributions (and in the official deployment tooling) especially important IMHO.

mowies:
I think at this point, bringing this topic up at a SIG meeting plus your mentioned threat analysis are probably a good idea.

TylerHelmuth:
Outside of the security discussion, I'm more concerned about adding a component to contrib that won't work on all the platforms/architectures we currently support. Rule 8 of the [distribution criteria](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions) states

```
8. Must include the following assets except where the specific purpose of the distribution is naturally associated with a subset of these assets. In such cases, it should be clearly stated which assets are skipped and why.  Additional assets may be included if the maintainers agree:
    - Binaries for linux_amd64, linux_arm64, windows_amd64 and darwin_arm64
    - linux_amd64 and linux_arm64 container images
    - Packages to be used with Linux distributions (RPM, deb), macOS (brew), Windows (MSI) for each distributed binary.
```

If a component makes the contrib distribution stop working for any of these assets then I don't think it can be included because contrib doesn't have any natural asset restrictions.

douglascamata:
> Outside of the security discussion, I'm more concerned about adding a component to contrib that won't work on all the platforms/architectures we currently support. Rule 8 of the [distribution criteria](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions?rgh-link-date=2026-07-16T19%3A29%3A29Z) states
> 
> ```
> 8. Must include the following assets except where the specific purpose of the distribution is naturally associated with a subset of these assets. In such cases, it should be clearly stated which assets are skipped and why.  Additional assets may be included if the maintainers agree:
>     - Binaries for linux_amd64, linux_arm64, windows_amd64 and darwin_arm64
>     - linux_amd64 and linux_arm64 container images
>     - Packages to be used with Linux distributions (RPM, deb), macOS (brew), Windows (MSI) for each distributed binary.
> ```
> 
> If a component makes the contrib distribution stop working for any of these assets then I don't think it can be included because contrib doesn't have any natural asset restrictions.

@TylerHelmuth: contrib will always have components that work only in certain environments and will be either skipped or cause a fatal error in other environments. For example, if you configure and enable the k8s receivers (`k8sclusterreceiver`, `k8seventsreceiver`, or `k8sobjectsreceiver`) outside of k8s, or any of the Windows receivers (`windowseventlogreceiver`, `windowsperfcountresreceiver`, or `windowsservicereiver` outside of Windows, or `dockerstatsreceiver` in a machine without Docker installed or supported at all, etc.

It's fine having unsupported platforms for a given components. We just gotta be careful with the behavior that I was asking about: "what happens if I enable this component on an unsupported platform? should it be a fatal error or warning?". If the component is present in the distro and not enabled/configured, the Collector MUST be able to start without issues.

TylerHelmuth:
@douglascamata I agree, but those components don't prevent the collector binary from starting. Based on the conversion in the PR it sounded like the profiler would. Please correct me if I'm misunderstanding.

douglascamata:
> @douglascamata I agree, but those components don't prevent the collector binary from starting. Based on the conversion in the PR it sounded like the profiler would. Please correct me if I'm misunderstanding.

Happy to clarify, @TylerHelmuth: according to https://github.com/open-telemetry/opentelemetry-collector-releases/pull/1553#issuecomment-4978909148, the eBPF profiler receiver will **not** prevent the Collector from starting in an unsupported platform, as long as it's **not** part of the configuration. 

If the eBPF profiler receiver is present in the configuration and the Collector is starting in an unsupported platform, the bootstrap of the component will error out and cause the Collector to exit.

From what I know, this matches the behavior of all those components I mentioned.

tl;dr: the Collector will only not start in unsupported platforms when the eBPF profiler receiver is configured. Its mere presence in the distribution is harmless.

dmathieu:
I don't know about this either. The profiler has its own distribution because it was always meant to run separately from a cluster's main collector.

In most common architectures, the main collector is a deployment. The profiler needs to be a daemonset.
The recommended approach is therefore to have 2 separate collectors. One that runs the profiler only, and the other as the main cluster's collector.
The profiler sends its data to the main collector, which sends it to any vendor.

To better help with that recommendation, and to avoid misleading users, I think it'd be better not to include the profiler into contrib. When you want to use, you setup a specific collector with the profiler distribution.

atoulme:
I think adding the ebpf profiler to this distro does a few things for me as a maintainer, and helps users. Here is how I see it:
* It helps others know the ebpf profiler can be mixed and matched with other components. For example, that obi conflict is interesting. The fact the ebpf profiler receiver fails to compile for Windows is also important to note.
* It lets users who want to use the ebpf profiler use one build, not two. This is important to users who don't want to have to scan multiple binaries.
* Yes, users can also make their own collector. But right now, they might not go that far to try it out. Having contrib with this receiver helps them get started faster, and you'll get more eyeballs that way.

dmathieu:
> Yes, users can also make their own collector.

They don't need to. There is already an ebpf distribution maintained by us.

Most of those reasons seem to be "we want to test at runtime". Is that the right way to do so though?

dmathieu:
Another thing to note is that ebpf-profiler isn't even alpha (only the profiles data model is).
Including it in contrib now would kinda break the "stable first" premise we're leaning towards.

One thing the profiling SIG mentioned before was to move the receiver code into collector-contrib, and keep the profiler repo only for the library.
That assumes much better stability that what we have now though.

PR #1567, opened 2026-07-22
chore(contrib): pin ebpf version to v0.21.0

mowies:
#### Description
This pin needed to happen because of a version incompatiblity between OBI and eBPF v0.22.0.
OBI ships with eBPF v0.20.0 and is not compatible with v0.22.0. Collector contrib has eBPF v0.22.0 as a transitive dependency and so during building, Go decides to use the higher version of the 2 which results in build error that can be see [here](https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/29881624893/job/88854526021?pr=1566).

#### Authorship


douglascamata (approved):

mowies:
@MrAlias maybe to ping you here as well. This can be reverted when OBI bump its eBPF dependency to >v0.22.0

PR #1569, opened 2026-07-27
[chore] Start Docker daemon on Windows release runners

### LINKED ISSUE #1495: Issues with windows runners during release
It seems that during releases, the windows runners sometimes have issues with docker. From what I could gather, the docker daemon is not properly started and then image builds fail as a result.

There is an upstream issue related to this: https://github.com/actions/runner-images/issues/13729

GitHub provided a workaround, which is to manually start the docker daemon service as one of the first steps in the pipeline:
https://github.com/actions/runner-images/issues/13729#issuecomment-4183956802

pujitha24:
Motivation:
During releases, the release-windows jobs (release-core, release-otlp,
release-contrib, release-k8s) sometimes fail because the Docker daemon
on GitHub-hosted Windows runners is not running when the workflow
tries to log in to Docker and build/push container images. This is a
known upstream GitHub Actions runner-images issue where the Docker
Windows service occasionally fails to come up cleanly on runner boot:
https://github.com/actions/runner-images/issues/13729

GitHub's own investigation into that upstream issue found the failure
occurs randomly at runner boot and is related to how the Docker
service creates its Hyper-V virtual switch; they have not identified a
root-cause fix, but published a workaround that (re)starts the Docker
service explicitly before it's needed, which they report eliminates
the failure in their testing. This change applies that same workaround
to reduce the likelihood of hitting this failure during our releases;
it is a mitigation for a randomly occurring runner issue, not a
guaranteed fix for the underlying flakiness.

Approach:
Add a "Start Docker daemon (Windows)" step, gated on
`runner.os == 'Windows'`, as the first step (right after checkout) in
both the `prepare` and `release` jobs of the reusable
base-release.yaml workflow, since both jobs perform Docker logins and
goreleaser-driven Docker image builds when running on the Windows
runner used for release-windows jobs. The step reuses the exact
pattern already established in base-ci-goreleaser.yaml
("Start Docker daemon (Windows)" / `shell: powershell` /
`run: Start-Service docker`) for consistency with this repo's existing
convention, rather than introducing a new pattern.

Validation:
- Verified YAML validity: python3 -c "import yaml; yaml.safe_load(open('.github/workflows/base-release.yaml'))"
- Ran `actionlint .github/workflows/base-release.yaml`; the only
  reported issues are pre-existing shellcheck false positives (present
  before this change too) caused by actionlint assuming a bash shell
  for other Windows PowerShell steps in this file that lack an
  explicit `shell:` key; the new steps declare `shell: powershell`
  explicitly and produce no lint output.
- This change only affects a GitHub Actions workflow file used during
  the release process (windows-2022 runners), so it cannot be
  exercised with a local Go build/test; there is no unit-testable code
  path for a workflow YAML change like this.

Fixes #1495

Assisted-by: Claude Sonnet 5
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

---
AI assistance: this change was drafted with Claude Code.


pujitha24:
The only actionable item I found was the failing `changelog` check — this repo requires a `.chloggen/*.yaml` entry for non-chore PRs, which I hadn't added. I added `.chloggen/start-docker-daemon-windows-release.yaml` (`change_type: bug_fix`, `component: release`, referencing #1495) describing this as a mitigation for the intermittent Windows Docker-daemon-not-running issue, consistent with the PR description. `make chlog-validate` passes locally.

The Copilot review left only a non-actionable summary (no requested changes), and I didn't find any other unanswered human feedback on this PR.

pujitha24:
This has been green and rebased for a couple weeks now — happy to make any changes if something would help move review along.

.chloggen/start-docker-daemon-windows-release.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
```
mowies: Please remove the changelog entry as this is just a non-user-facing pipeline change

mowies (commented):
please remove the changelog entry, otherwise LGTM 👍🏼

pujitha24:
Removed the .chloggen entry — makes sense, this is pipeline-only. One thing to flag: the changelog workflow's "Ensure ./.chloggen/*.yaml addition(s)" check will likely fail again now since the PR title doesn't have `[chore]` and there's no `Skip Changelog` label; I don't have permissions to add the label, so if you can add it (or let me know if you'd rather I retitle) that'd unblock it.

mowies (approved):

PR #1593, opened 2026-08-09
[chore] Retry GHCR login on failure in release workflows

### LINKED ISSUE #1496: GHCR logins failed multiple times during v0.152.0 release
The docker/login-action failed multiple times on different release workflows during the v0.152.0 release. The action just ran into timeout without any other error being thrown. I suspect that we run into some API rate limit with the release workflows all running at the same time.

For now, I think the best option would be to use some retry logic around the login-action (e.g.https://github.com/nick-fields/retry) and to use [max-parallel](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#defining-the-maximum-number-of-concurrent-jobs) on some/all of the release matrices to cut down on concurrency a bit.

pujitha24:
#### Description

The `docker/login-action` step that logs into GHCR (`ghcr.io`) failed multiple times
during the v0.152.0 release, timing out with no other error. The maintainer who filed
the issue suspected GHCR API rate limiting from many release workflow jobs running
concurrently, and suggested wrapping the login in retry logic (e.g.
[nick-fields/retry](https://github.com/nick-fields/retry)).

This change wraps the "Login to GitHub Package Registry" step in `base-release.yaml`
(both the `prepare` and `release` jobs) and `base-binary-release.yaml` with
`nick-fields/retry`, retrying the underlying `docker login` up to 5 times with a 15s
wait and a 2 minute per-attempt timeout. The separate Docker Hub login step is left
unchanged since the reported failure is specific to GHCR.

Validation:
- `actionlint .github/workflows/base-release.yaml .github/workflows/base-binary-release.yaml`
  before and after the change: no new findings (only pre-existing, unrelated shellcheck
  warnings on other `run:` blocks, present before this change too).
- Verified `nick-fields/retry@ad984534de44a9489a53aefd81eb77f87c70dc60` is the exact
  commit for the `v4.0.0` tag via the GitHub API.
- Smoke-tested the underlying `echo "$TOKEN" | docker login ghcr.io -u "$OWNER"
  --password-stdin` shape locally against a fake `docker` script that fails twice then
  succeeds, confirming the retry-then-succeed flow and that piping the token via stdin
  works as `docker/login-action` itself does internally.
- Not tested: an actual GHCR rate-limit event, or a live GitHub Actions run (no release
  secrets or hosted runner available here). This also was not dry-run on a
  `windows-2022` runner (used for the Windows GOOS release jobs) — `shell: bash` is a
  normal, supported option on GitHub's Windows runners and the prior `docker/login-action`
  step already required a working `docker` CLI in the same job, so the raw `docker login`
  invocation is expected to behave the same way there, but this specific combination
  with `nick-fields/retry` has not been exercised on Windows.

#### Link to tracking issue

Fixes https://github.com/open-telemetry/opentelemetry-collector-releases/issues/1496

#### Authorship

- [ ] I, a human, wrote this pull request description myself.

---
AI assistance: this change was drafted with Claude Code.

Fixes #1496


.chloggen/retry_ghcr_login.yaml
mowies, line 1: A changelog entry is not required for pipeline changes since they are not user facing, please remove it.

.github/workflows/base-binary-release.yaml
```diff
@@ -126,11 +126,13 @@ jobs:
...
+        uses: nick-fields/retry@ad984534de44a9489a53aefd81eb77f87c70dc60 # v4.0.0
         with:
-          registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.GITHUB_TOKEN }}
+          timeout_minutes: 2
+          max_attempts: 5
+          retry_wait_seconds: 15
+          shell: bash
+          command: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u "${{ github.repository_owner }}" --password-stdin
```
mowies: please put the github token into an env variable and then reference it in the command. this way, we don't need to rely on github actions to mask the plain token in the logs

mowies (approved):

mowies:
Thanks for your contribution!

pujitha24:
@mowies — a separate ask: I'm applying for OpenTelemetry org membership and it needs two approver sponsors from different companies. You reviewed and approved #1569 and #1593 here. Would you consider sponsoring? My merged work covers collector-releases, opentelemetry-operator, collector-contrib and go-contrib. No worries at all if you'd rather not.

PR #1615, opened 2026-08-19
Remove stale cilium/ebpf dependency override

MrAlias:
#### Description

Remove the temporary `github.com/cilium/ebpf v0.21.0` replacement from the `otelcol-contrib` manifest while retaining the local OBI source replacement.

OBI v0.11.0 requires cilium/ebpf v0.22.0 and uses the `CollectionOptions.Cache` API introduced in that version. The stale override forces v0.21.0 and causes the collector build to fail because that API is not available. Without the override, Go resolves cilium/ebpf v0.22.0 and OBI v0.11.0 compiles successfully.

#### Link to tracking issue

Supersedes #1604

#### Authorship

### DIFF
--- a/.chloggen/remove-obi-ebpf-override.yaml
+++ b/.chloggen/remove-obi-ebpf-override.yaml
@@ -1,0 +1,6 @@
+change_type: bug_fix
+component: otelcol-contrib
+note: Remove the obsolete cilium/ebpf v0.21.0 override so OBI v0.11.0 builds with v0.22.0.
+issues: [1604]
+subtext:
+change_logs: [user]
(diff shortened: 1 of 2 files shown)


.chloggen/remove-obi-ebpf-override.yaml
```diff
@@ -0,0 +1,6 @@
+change_type: bug_fix
```
songy23: I would call it a "enhancement" with go.opentelemetry.io/obi bumped to v0.11.0. Removing the obsolete cilium/ebpf v0.21.0 override has no visible impact to end users AFAIS

songy23 (approved):

PR #1633, opened 2026-09-02
[chore] [ci] don't fail Windows docker-tests on collector stderr output

dmitryax:
The Windows `docker-tests` step runs under `shell: powershell` (Windows PowerShell 5.1) with `$ErrorActionPreference = 'Stop'`. The collector writes its startup logs to stderr, and capturing them with `*>` promotes that native-command output to a terminating `NativeCommandError`, so the step exits 1 even though the container starts fine and the `"Everything is ready."` check is never reached.

This broke at #1627, which added `shell: powershell` to the step. Before that it ran under `pwsh` (PowerShell 7), which doesn't promote native stderr to a terminating error. It has failed on `main` for Core, OTLP, and k8s Windows docker-tests since; every PR inherits the red checks.

## Fix
Set `$ErrorActionPreference = 'Continue'` for the log capture and use `2>&1 | Out-File`, so stderr is captured without being treated as fatal. The readiness check is unchanged.

- [ ] I, a human, wrote this pull request description myself.

### DIFF
--- a/.github/workflows/base-ci-goreleaser.yaml
+++ b/.github/workflows/base-ci-goreleaser.yaml
@@ -428,7 +428,13 @@ jobs:
           echo "Sleeping for a bit..."
           sleep 10
           echo "Checking logs for correct startup..."
-          docker logs $env:INPUTS_DISTRIBUTION *> "docker-logs-${env:INPUTS_DISTRIBUTION}.log"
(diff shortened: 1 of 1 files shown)


dmitryax:
Superseded: reopening from my fork branch (dmitryax:fix/windows-docker-logs-stderr) instead of an upstream branch.

